Issue #23, opened 2023-07-28
🔧 Investigate CGO Usage for WASMVM


gluax:
## 🔧 Chore

We may possibly not need CGO for the [WASMVM](https://github.com/CosmWasm/wasmvm) usage.

## Motivation

If we don't need CGO the build process would be a lot easier and simpler for people to work with. Could also simplify the CI. The linked repo has instructions on not using CGO. Unsure if this means we wouldn't need it in the end or not.

gluax:
Priority change due to affecting how we deal with the WASM shared object library file for upgrades.

mariocao:
Timeboxed to half a day

PR #27, opened 2023-08-15
chore: Add Makefile for build and lint

### LINKED ISSUE #22: ✨ Migrate From Ignite
## ✨ Feature

To migrate away from `ignite`

## Motivation

A lot of other `cosmos` based chains have migrated away from `ignite` for building, testing, etc. There's likely some reason for this.

## Implementation

Let's decide whether to use `just` or `make`. Either way, we can likely copy a large amount of this from other projects. Then we can document them in the appropriate markdown files.

hacheigriega:
## Motivation

Add a Makefile for building binaries and linting the codebase.

## Related PRs and Issues

Addresses Issue #22 :  We can continue using Ignite, but a Makefile will allow us to write custom commands. Using Ignite also makes command running process opaque.

Also related to Issue #23 : CGO is enabled for build and install processes for now. Further investigation is needed to see if this brings any issues like indeterminacy.

Closes #22


mariocao (approved):
Looks good! You forgot to sign the first 2 commits :)

hacheigriega:
@mariocao Is it ok to force push so that all commits are signed and follow conventional commits guide?

gluax:
@hacheigriega If that's after a rebase or amend should be good!

Also, if you could make the changes I suggested before merging and mark the conversations as resolved that would be great!

PR #35, opened 2023-08-24
feat: x/storage module implementation

### LINKED ISSUE #33: ✨ Build a skeletal module x/storage
## ✨ Feature

I am proposing a simple implementation of module called `x/storage`, which would accomplish a few things:

- Replace `x/sedachain`, whose structure is confusing and purpose unclear. Instead, `x/storage` would serve as a skeletal module to facilitate future development of a new module. Even though its functionality would be very limited, it would nonetheless involve the usual groundwork for module development such as setting up the general module code structure and Protobuf generation process.
- The `x/storage` module can also be used to test storing binaries on chain through a module store. We can check if a module transaction is affected by the same size limitation that affects a CosmWasm transaction.

hacheigriega:
This PR implements a simple module for storing a Wasm or gzip file.

## Related PRs and Issues

Closes #33


Thomasvdam (commented):
Wasn't sure where to put it in code so I'll leave it here.

Does it make sense to add some additional fields already, even though we might not be using them right away? I'm thinking of
- type: is it a dr execution binary, dr tally binary, executor/relayer binary, etc.
- Some sort of block height timestamp: for when we want to implement pruning.

hacheigriega:
@Thomasvdam Yes I agree, and I'm actually working on that on another branch. I was thinking we could push this PR first since it's getting quite long. This PR simply sets up a skeletal module without much internal logic.

x/storage/keeper/msg_server.go
```diff
@@ -0,0 +1,59 @@
...
+var _ types.MsgServer = msgServer{}
+
+func (k msgServer) Store(goCtx context.Context, msg *types.MsgStore) (*types.MsgStoreResponse, error) {
+	ctx := sdk.UnwrapSDKContext(goCtx)
+
+	data := msg.Data
+
+	var err error
+	if ioutils.IsGzip(data) {
+		data, err = ioutils.Uncompress(data, int64(800*1024))
```
FranklinWaller, line 35: Just to be sure. We are not susceptible to zip bombing?

FranklinWaller (approved):

mariocao (approved):
Looks good! :)

x/storage/types/events.go
```diff
@@ -0,0 +1,7 @@
+package types
+
+const (
+	EventTypeStore = "store"
+
+	AttributeHash = "hash"
+)
```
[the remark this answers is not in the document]
hacheigriega: I'm all for defining a standard. Thanks for the proposal, and I'll make sure to address it in the new PR that will contain a complete implementation of wasm-storage module.

x/storage/keeper/msg_server.go
hacheigriega: Thanks for the heads up. Will make sure of this in the new PR that will contain full implemenetation of the module.

PR #39, opened 2023-08-30
chore: update default genesis

hacheigriega:
## Explanation of Changes

This PR adds wrappers around certain Cosmos modules so that we can overwrite their default genesis states to fit our needs.

## Testing

Generated a `genesis.json` using the `init` command to check that `community_tax` is now `0.000000000000000000` and that `bond_denom` and `mint_denom` are now `aseda`.


gluax (approved):
LGTM!

Just want to make sure we tested generating a new `genesis.json` and we can start without modifying anything in that file now @hacheigriega?

PR #40, opened 2023-09-06
feat: Wasm storage for Data Request & Overlay binaries

### LINKED ISSUE #41: ✨Wasm storage for Data Request & Overlay binaries
## ✨ Feature

Replace the skeletal module `x/storage` to create a new module called `x/wasm-storage` for storing all Wasm binaries used in the SEDA protocol. 

## Implementation

This new module's store will contain two separate keys for storing Data Request Wasm binaries and Overlay Wasm binaries. 

The Wasm binaries keyed by their hashes will be stored in compressed bytecode format along with the specific type information. Note that Data Request Wasm binaries are further categorized into Data Request Wasm and Tally Wasm, while Overlay Wasm binaries are further categorized into Data Request Executor Wasm and Relayer Wasm.

```
0x00 | Hash(DataRequestWasm) -> Wasm{}
0x01 | Hash(OverlayWasm) -> Wasm{}
```

The module's transactions for storing Wasms will look as follows:
```
$ seda-chaind tx wasm-storage store-data-request-wasm <filename> --wasm-type data-request
$ seda-chaind tx wasm-storage store-data-request-wasm <filename> --wasm-type tally

$ seda-chaind tx wasm-storage store-overlay-wasm <filename> --wasm-type data-request-executor
$ seda-chaind tx wasm-storage store-overlay-wasm <filename> --wasm-type relayer
```

Make sure to include ways to query a binary using its hash and list hashes of all binaries. 
```
$ seda-chaind query wasm-storage data-request-wasm <wasm_hash>
$ seda-chaind query wasm-storage overlay-wasm <wasm_hash>

$ seda-chaind query wasm-storage all-data-request-wasms
$ seda-chaind query wasm-storage all-overlay-wasms
```

hacheigriega:
## ✨ Feature

The new `x/wasm-storage` module provides a place where Data Request Issuers can store their Data Request Wasm binaries, which can then be fetched by Overlay nodes. It also stores Overlay Wasm binaries in a separate bucket. Another PR will put these Overlay Wasms under governance control.

## Implementation

As discussed in the issue (#41), there are two separate buckets for Data Request Wasms and Overlay Wasms.
```
0x00 | Hash(DataRequestWasm) -> Wasm{}
0x01 | Hash(OverlayWasm) -> Wasm{}
```

where `Wasm{}` struct type looks as follows:
```
type Wasm struct {
	Hash     []byte
	Bytecode []byte
	WasmType WasmType
}
```

Transactions on this module looks as follows:
```
$ seda-chaind tx wasm-storage store-data-request-wasm <filename> --wasm-type data-request
$ seda-chaind tx wasm-storage store-data-request-wasm <filename> --wasm-type tally

$ seda-chaind tx wasm-storage store-overlay-wasm <filename> --wasm-type data-request-executor
$ seda-chaind tx wasm-storage store-overlay-wasm <filename> --wasm-type relayer
```

Queries can be made using hashes provided in transaction results. There are also endpoints for listing all hashes and Wasm types in the store:
```
$ seda-chaind query wasm-storage data-request-wasm <hash>
$ seda-chaind query wasm-storage overlay-wasm <hash>


$ seda-chaind query wasm-storage list-data-request-wasms
$ seda-chaind query wasm-storage list-overlay-wasms
```

Please note that gas adjustment flag may have to be set to about 1.2 since simulation often underestimates gas usage. That is, add the flags `--gas auto --gas-adjustment 1.2` to transactions. Also note that following the CosmWasm standard we set the Wasm size limit at 819.2 KB in its uncompressed state. This limit is also used during unzip process to prevent zip bombing.

## Testing

Testing will be addressed in a separate PR. Checked that transaction and query endpoints work as expected with Wasm files of varying sizes.

## Related PRs and Issues

Closes #41


proto/sedachain/wasm_storage/query.proto
```diff
@@ -0,0 +1,65 @@
...
+  }
+
+  rpc OverlayWasms(QueryOverlayWasmsRequest)
+      returns (QueryOverlayWasmsResponse) {
+    option (google.api.http).get = "/seda-chain/wasm-storage/overlay_wasms";
+  }
+}
+
+message QueryDataRequestWasmRequest {
+  string hash = 1 [ (gogoproto.moretags) = "yaml:\"hash\"" ];
```
gluax, line 38: Possibly a dumb question:

What's the `yaml` specifier for here in these messages? Isn't `json` more standard for `rpc` things?

x/wasm-storage/types/wasm.go
```diff
@@ -0,0 +1,51 @@
...
+	if len(s) > MaxWasmSize {
+		return fmt.Errorf("Wasm code cannot be longer than %d bytes", MaxWasmSize)
+	}
+	return nil
+}
+
+// NewWasm constructs a new Wasm object given bytecode and Wasm type.
+// It panics if it fails to compute hash of bytecode.
+func NewWasm(bytecode []byte, wasmType WasmType) *Wasm {
+	hash := crypto.Keccak256(bytecode)
```
gluax, line 26: Ignore this marker. It's so we know what hash algo to use on the overlay side :).

x/wasm-storage/keeper/msg_server.go
```diff
@@ -0,0 +1,85 @@
...
+}
+
+var _ types.MsgServer = msgServer{}
+
+func (k msgServer) StoreDataRequestWasm(goCtx context.Context, msg *types.MsgStoreDataRequestWasm) (*types.MsgStoreDataRequestWasmResponse, error) {
+	ctx := sdk.UnwrapSDKContext(goCtx)
+
+	unzipped := unzipWasm(msg.Wasm)
+	wasm := types.NewWasm(unzipped, msg.WasmType)
+	if k.Keeper.HasDataRequestWasm(ctx, wasm) {
```
gluax, line 31: I think the goal is to be able to overwrite a hash since we are permissionless yes @mariocao @FranklinWaller?

x/wasm-storage/types/keys.go
```diff
@@ -0,0 +1,45 @@
+package types
+
+import (
+	"time"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+)
+
+const (
```
gluax, line 9: It may be a stylistic choice on my end, but I'm unsure if others would agree; however, I noticed we have some configured constants in many different files. I think it's easier to find/edit if they are usually in the same place, like a `consts.go` file, and then labeled appropriately in there. Food for thought, not a hard ask.

gluax (commented):
It looks good to me, waiting on approval though to double-check the overwriting part :).

x/wasm-storage/types/wasm.go
```diff
@@ -0,0 +1,51 @@
...
+	if hash == nil {
+		panic("failed to compute hash")
+	}
+
+	return &Wasm{
+		Hash:     hash,
+		Bytecode: bytecode,
```
FranklinWaller, line 33: Maybe for future usage we can add the block it was added to? So we don't have to migrate our state if we ever introduce pruning

x/wasm-storage/types/wasm.go
```diff
@@ -0,0 +1,51 @@
...
+
+import (
+	fmt "fmt"
+	"strings"
+
+	"github.com/hyperledger/burrow/crypto"
+)
+
+// MaxWasmSize is the maximum size of Wasm bytecode.
+const MaxWasmSize = 800 * 1024
```
FranklinWaller, line 11: I'm not sure so correct me if i'm wrong but isn't this better to have as a genesis config thing? So we can always upgrade it through a gov decision

x/wasm-storage/keeper/msg_server.go
```diff
@@ -0,0 +1,85 @@
...
+		Hash: hashString,
+	}, nil
+}
+
+// unzipWasm unzips a gzipped Wasm into
+func unzipWasm(wasm []byte) []byte {
+	var unzipped []byte
+	var err error
+	if ioutils.IsGzip(wasm) {
+		unzipped, err = ioutils.Uncompress(wasm, types.MaxWasmSize)
```
FranklinWaller, line 79: I don't remember if i asked this question earlier but is this safe against zipbomb attacks?

x/wasm-storage/keeper/msg_server.go
FranklinWaller: The hash is based on the bytes so overwriting is not a thing right

hacheigriega: @FranklinWaller Oh the `Wasm` struct has the field `WasmType`, which could be overwritten.

hacheigriega: We could just get rid of `WasmType` if it's unnecessary to store this info on chain.

x/wasm-storage/types/keys.go
hacheigriega: That's a reasonable suggestion, but I'm just following the standards of Cosmos projects here. (putting all key-related stuff in `keys.go`)

x/wasm-storage/types/wasm.go
hacheigriega: I agree. Not sure if we're going with block number-based vs block time-based pruning but I'll add block number for now.

x/wasm-storage/types/wasm.go
hacheigriega: Yea that was actually my initial idea but got lazy 😋

x/wasm-storage/keeper/msg_server.go
hacheigriega: Yes I checked that. `MaxWasmSize` should prevent that.

x/wasm-storage/keeper/msg_server.go
gluax: Yea, that's the only thing I was wondering about. I should have been more specific. My bad!

Eventually, if/when we add metadata too, that could be an issue... maybe rather than storing metadata and that stuff on the chain, it could emit an event or something for the indexer to grab with that info? i.e. we still take it as args, but we don't store it just give it to indexer in some way.

proto/sedachain/wasm_storage/query.proto
hacheigriega: Not dumb at all.. probably for query results? Copy-pasting went too far there. I removed them.

x/wasm-storage/types/wasm.go
hacheigriega: Acutally I added timestamp of when the object was added because I think pruning should be based on timestamps (not important anyways right now)

FranklinWaller: To be honest, i think block number is probably better because it's more consistent. Timestamps are not really to be trusted. But maybe it doesn't matter in this use case. We can resolve this one for now.

x/wasm-storage/types/wasm.go
FranklinWaller: Hahaha, no worries, atleast it's noted

FranklinWaller (approved):

hacheigriega:
@gluax I'll merge if you approve :)

x/wasm-storage/keeper/msg_server.go
Thomasvdam: Late to the party but I think the idea is to have the metadata be indexer only. So as long as you have the private key used to upload the bytes you can edit the metadata through the explorer.

PR #46, opened 2023-09-11
feat: gov-controlled Overlay Wasms

### LINKED ISSUE #44: ✨ Add governance control of Overlay Wasm binaries
## ✨ Feature

Overlay Wasm binaries, which include Data Request Executor and Relayer Wasm binaries, should be under control of governance process. 

Implementation details remain undetermined, but we need to keep track of previous, current, and next binaries for each type.

hacheigriega:
## Explanation of Changes

This PR replaces the existing transaction endpoint for sending an `sdk.Msg` for storing an Overlay Wasm binary with a governance proposal submission endpoint that wraps this message. As a result, any updates in the Overlay Wasm bucket now require going through governance voting process. By default settings, a proposal is passed with 1/3 of voters weighted by their stakes voting and a majority yes.

## Testing

Manually tested in local setting. The local setup script was updated so that voting period is short (30 seconds).

After setting up the chain with `scripts/local_setup.sh`, run the following commands to submit a proposal and vote yes.
```
$ seda-chaind tx wasm-storage submit-proposal store-overlay-wasm <wasm_file> --wasm-type relayer --title <title> --summary <summary> --deposit 10000000aseda --from <test_key_addr> --keyring-backend test --gas auto --gas-adjustment 1.2 -y

$ seda-chaind tx gov vote 1 yes --from <test_key_addr> --keyring-backend test --gas auto --gas-adjustment 1.2 -y

# after 30 seconds, the Wasm should appear on the list
$ seda-chaind query wasm-storage list-overlay-wasms
```

## Related PRs and Issues

Closes #44

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -566,6 +566,7 @@ func NewApp(
 	app.WasmStorageKeeper = *wasmstoragekeeper.NewKeeper(
 		appCodec,
 		keys[wasmstoragetypes.StoreKey],
+		authtypes.NewModuleAddress(govtypes.ModuleName).String(),
 	)
 	wasmStorageModule := wasmstorage.NewAppModule(appCodec, app.WasmStorageKeeper, app.AccountKeeper, app.BankKeeper)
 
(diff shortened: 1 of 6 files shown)


JeancarloBarrios (approved):
LGTM

JeancarloBarrios:
I think maybe in a future PR, it would be nice, and I can do it as well to have a spec for this module.

PR #50, opened 2023-09-18
feat(wasm-storage): upgrade emitevent to emit typed event

JeancarloBarrios:
## Motivation

EmitEvent is deprecated therefore I upgraded to EmitTypeEvent


proto/sedachain/wasm_storage/v1/events.proto
```diff
@@ -0,0 +1,21 @@
...
+
+option go_package = "github.com/sedaprotocol/seda-chain/x/wasm-storage/types";
+
+
+message EventDataRequestWasm {
+  string hash = 1;
+  WasmType wasm_type = 2;
+
+}
+
```
JeancarloBarrios, line 17: An alternative could be: 
`message EventWasm {
  string hash = 1;
  WasmType wasm_type = 2;
   RequestType request_type = 3;

}`

@hacheigriega what do you think?

JeancarloBarrios: The module would be shown in the proto-type. This is an example of group modue event
image

proto/sedachain/wasm_storage/v1/wasm_storage.proto
```diff
@@ -28,4 +28,14 @@ enum WasmType {
       [ (gogoproto.enumvalue_customname) = "WasmTypeRelayer" ];
 }
 
+enum RequestType {
```
hacheigriega, line 31: Oops I should have pointed out earlier, but this is a misnomer because the wasm type is either Data Request or Overlay. I don't think we will need `RequestType` since these two types can be determined by tx type. Yes I know it's a little confusing 😂

proto/sedachain/wasm_storage/v1/events.proto
```diff
@@ -0,0 +1,17 @@
...
+import "sedachain/wasm_storage/v1/wasm_storage.proto";
+
+option go_package = "github.com/sedaprotocol/seda-chain/x/wasm-storage/types";
+
+
+message EventStoreWasm {
+  string hash = 1;
+  WasmType wasm_type = 2;
+  RequestType request_type = 3;
+
```
hacheigriega, line 16: Try linting using `make proto-lint`. Feel free to change the linter if you dislike it btw..

hacheigriega (changes requested):

proto/sedachain/wasm_storage/v1/wasm_storage.proto
JeancarloBarrios: it is to identify the event. But I'll change back to two events.

hacheigriega (approved):

PR #51, opened 2023-09-19
feat(wasm-storage): add unit test for msg server

JeancarloBarrios:
Added test for msg server and querier


x/wasm-storage/keeper/msg_server.go
```diff
@@ -76,14 +82,15 @@ func (k msgServer) StoreOverlayWasm(goCtx context.Context, msg *types.MsgStoreOv
...
+func unzipWasm(wasm []byte) ([]byte, error) {
 	var unzipped []byte
 	var err error
-	if ioutils.IsGzip(wasm) {
-		unzipped, err = ioutils.Uncompress(wasm, types.MaxWasmSize)
-		if err != nil {
-			panic(err)
-		}
+	if !ioutils.IsGzip(wasm) {
+		return nil, fmt.Errorf("wasm is not gzip compressed")
```
JeancarloBarrios, line 89: Change this so it returns an error when the wasm was not compressed, to fix the behavior "non gzip wasm files would be hash as nil"

hacheigriega: nice catch - tyty

hacheigriega (approved):
LGTM. Thank you for writing all these tests.

x/wasm-storage/keeper/msg_server.go
hacheigriega: nice catch.. tyty

PR #52, opened 2023-09-20
chore: Generating Protobuf code in Rust

hacheigriega:
## Motivation

This PR sets up tools for generating Rust bindings for the chain's Protobuf schema.

## Explanation of Changes

Added a new Makefile command `make proto-gen-rust` for generating Rust bindings based on the proto files. It uses [Prost](https://github.com/tokio-rs/prost) for general code generation and [Tonic](https://github.com/hyperium/tonic) for gRPC code generation.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -119,6 +119,10 @@ proto-gen:
 	@echo "Generating Protobuf files"
 	@$(protoImage) sh ./scripts/protocgen.sh
 
+proto-gen-rust:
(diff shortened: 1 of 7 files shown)


FranklinWaller:
This is perfect! I think only we should move this to the overlay side so you guys don't have dangling rust files.

gluax:
Should be moved to overlay side and we should use Prost-build as a build dependency and have a build.rs file handle this.

PR #57, opened 2023-09-26
chore: Fix event attributes

hacheigriega:
This PR adds missing proto versions from event attributes and fixes string fields format.


JeancarloBarrios (approved):

Thomasvdam (approved):
Looks good, I get `sedachain.wasm_storage.v1.EventStoreDataRequestWasm` as event type.

Lets leave the quotes issue for now, I'll just strip escaped double quotes from the start and end of the string if they're present.

Lastly, is it easy/advisable to emit the WASM bytes in the event as well? I could grab them from the `message`, but it's a lot easier if it's just another attribute in the event.

As a nice to have is it possible to capture the format for the package in the proto files in a linting rule? Only if it's easy to do, don't bother if it's a lot of work. :)

PR #60, opened 2023-09-28
fix: return error when Wasm uncompression fails

hacheigriega:
This PR fixes a bug that was introduced by an incorrect error handling. It also fixes message server unit tests and adds a few more cases.


x/wasm-storage/keeper/msg_server_test.go
```diff
@@ -11,9 +11,15 @@ import (
 )
 
 func (s *KeeperTestSuite) TestStoreDataRequestWasm() {
-	wasm, err := os.ReadFile("test_utils/hello-world.wasm")
+	regWasm, err := os.ReadFile("test_utils/hello-world.wasm")
+	s.Require().NoError(err)
+	regWasmZipped, err := ioutils.GzipIt(regWasm)
+	s.Require().NoError(err)
+
+	oversizedWasm, err := os.ReadFile("test_utils/simple-dr.wasm")
```
Thomasvdam, line 19: Maybe we should name the file `oversized.wasm`. Right now it feels a bit silly that a 'simple' WASM binary is considered oversized. 😬

Thomasvdam (approved):

x/wasm-storage/keeper/msg_server_test.go
```diff
@@ -28,32 +34,66 @@ func (s *KeeperTestSuite) TestStoreDataRequestWasm() {
...
 					WasmType: types.WasmTypeDataRequest,
 				}
 				_, err := s.msgSrvr.StoreDataRequestWasm(s.ctx, &input)
 				s.Require().Nil(err)
 			},
 			expErr:    true,
-			expErrMsg: "Data Request Wasm with given hash already exists",
+			expErrMsg: "data Request Wasm with given hash already exists",
+		},
+		// TO-DO: Add after migrating ValidateBasic logic
```
mariocao, line 64: Why commented?

Is the logic not already there?
https://github.com/sedaprotocol/seda-chain/blob/6a1f50d7e9e0187817f466fa3c15406037fb7ff5/x/wasm-storage/types/tx.go#L22

JeancarloBarrios (approved):

x/wasm-storage/keeper/msg_server_test.go
hacheigriega: Yes! But Cosmos is encouraging us to put that logic directly in `msg_server` starting with v0.50. See https://github.com/cosmos/cosmos-sdk/issues/15648

PR #137, opened 2023-12-19
Add Ability to Change max-wasm-size via Governance

### LINKED ISSUE #62: ✨ Parameterize max Wasm size in wasm-storage module
## ✨ Feature

Currently, the maximum Wasm size is set using a constant (~800kb). Using a module parameter for this value instead would allow us to adjust it through the governance process.

jim380:
## Overview
This PR attempts to achieve the followings:
- The option to override `max-wasm-size` at compile time
- The to change `max-wasm-size` via governance

Closes #62


x/wasm-storage/keeper/msg_server.go
```diff
@@ -173,3 +173,16 @@ func unzipWasm(wasm []byte) ([]byte, error) {
 	}
 	return unzipped, nil
 }
+
+func (m msgServer) UpdateParams(goCtx context.Context, req *types.MsgUpdateParams) (*types.MsgUpdateParamsResponse, error) {
+	if m.GetAuthority() != req.Authority {
+		return nil, fmt.Errorf("invalid authority; expected %s, got %s", m.GetAuthority(), req.Authority)
+	}
+
```
hacheigriega, line 181: v0.50 recommends that we validate message here

x/wasm-storage/keeper/msg_server.go
```diff
@@ -173,3 +173,16 @@ func unzipWasm(wasm []byte) ([]byte, error) {
...
+	ctx := sdk.UnwrapSDKContext(goCtx)
+	if err := m.SetParams(ctx, req.Params); err != nil {
```
hacheigriega, line 183: validate params before setting?

x/wasm-storage/types/tx.go
```diff
@@ -87,3 +87,15 @@ func (msg MsgInstantiateAndRegisterProxyContract) ValidateBasic() error {
 	}
 	return nil
 }
+
+func (m *MsgUpdateParams) ValidateBasic() error {
```
hacheigriega, line 91: maybe call it `Validate()` to avoid confusion? (since `ValidateBasic()` has been deprecated)

hacheigriega:
Closes #62, right?

What's the idea behind the compile time parameter? Just wondering why we would ever want that when the max wasm size is a chain parameter.

app/app.go
```diff
@@ -589,6 +590,12 @@ func NewApp(
 		panic(err)
 	}
 
+	if maxSize := os.Getenv("MAX_WASM_SIZE"); maxSize != "" {
```
hacheigriega, line 593: What's the idea behind the compile time parameter? Just wondering why we would ever want that when the max wasm size is a chain parameter.

hacheigriega: Perhaps just for testing purpose?

jim380: Because they might need to be used by `ValidateBasic()` or `Validate()` which is stateless.

jim380: But yeah if it was changed via governance afterwards, the binary wouldn't be able to pick that up without accessing state. For testing purposes for sure.

x/wasm-storage/keeper/msg_server.go
jim380: The msg (`MsgUpdateParams`) would've been validated in `Validate()`:
```golang
func (m *MsgUpdateParams) Validate() error {
	if _, err := sdk.AccAddressFromBech32(m.Authority); err != nil {
		return fmt.Errorf("invalid authority address: %s", err)
	}

	if err := m.Params.ValidateBasic(); err != nil {
		return err
	}

	return nil
}
```

hacheigriega (approved):

PR #140, opened 2023-12-27
Set Max Block gas in InitChainer

jim380:
## Overview
So that we won't have to manipulate the genesis every time.


hacheigriega:
Closing as we can just manipulate genesis every time

PR #209, opened 2024-03-07
Proposer sanity

DeshErBojhaa:
## Transaction sanity check

- Perform the custom sanity check
- Then call the Proposal -> Processor from the base app.

## What went well
- I am happy that I was able to do it in under 4 hours, not knowing about the latest abci++

## What went wrong
- Inaccurate time tracking due to switching between a full-time job and this.
- The majority of the allocated time went into researching abci++
- **That left no time for unit tests.**
- Bad error handling and could make smart ways to incorporate all the Proposal changes from any modules.

## Proposal for better evaluation of future candidates
- Sending the assignment over the weekend.


hacheigriega:
Thank you for the submission!!

PR #213, opened 2024-03-15
chore: upgrade dependencies &  make minor proto change for Keplr support

hacheigriega:
## Explanation of Changes

- The embedded field's name in `ClawbackContinuousVestingAccount` has been modified upon Keplr's request
- Updating dependencies


Thomasvdam (approved):
Just to be sure 😬 we tested that spinning up a node with this version while using a genesis that contains the old serialised account format?

hacheigriega:
@Thomasvdam Yes I tested that, and there was no issue. Will also test this on devnet before upgrading testnet.

hacheigriega:
Protobuf doesn't seem to care about the name of an embedded type at all when it comes to marshaling/unmarshaling.

mariocao (approved):

Issue #211, opened 2024-03-13
✨ State streaming plugin for indexing


Thomasvdam:
## ✨ Feature

As discussed internally we'd like to change the indexing approach to a push based model, where the chain takes the lead and pushes any important updates to the indexer for persistence and processing.

## Motivation

This allows us to do any bytes/proto parsing closer to the source and send JSON over the wire, which is easier for the indexer to process and reduces the need for libraries like CosmJS. Additionally this way the chain will push updates when they happen and we don't have to guess when certain states change on the indexer side.

## Implementation

Based on the SDK capabilities outlined in https://github.com/cosmos/cosmos-sdk/blob/v0.50.3/docs/architecture/adr-038-state-listening.md we feel that this plugin architecture is the best course of action.

This plugin would be responsible for parsing supported state changes, serialising them to JSON, and publishing them on an SQS queue.

PR #214, opened 2024-03-18
Add indexing plugin skeleton

Thomasvdam:
## Motivation

Initial skeleton to guide how future modules can be implemented, as well as SQS connectivity.

## Explanation of Changes

On the boilerplate level the main missing feature is logging. For some reason logs won't show up in the node logs no matter how I configure it.

## Testing

It's a handful but the following steps should help you verify that it works:

0. Build and initialise the chain.
1. Set up SQS emulator. I'd recommend https://github.com/Admiral-Piett/goaws (they provide a docker image) with the following config file:
```yaml
Local:
  Host: sqs
  Region: eu-west-2
  Queues:
    - Name: local-updates.fifo
```
2. Build the plugin with the dev build tag: `go build --tags dev -o build/plugin ./plugins/abci_indexing/plugin.go`.
3. Export the following env variables in a terminal session from which you will start the node.
```sh
export COSMOS_SDK_ABCI=$(readlink -f build/plugin)
export SQS_QUEUE_URL="http://localhost:4100/local-updates.fifo"
export SQS_ENDPOINT=http://localhost:4100
```
4. Edit the `State Streaming` section of the `app.toml` of the node:
```toml
# List of kv store keys to stream out via gRPC.
# The store key names MUST match the module's StoreKey name.
#
# Example:
# ["acc", "bank", "gov", "staking", "mint"[,...]]
# ["*"] to expose all keys.
keys = ["*"]

# The plugin name used for streaming via gRPC.
# Streaming is only enabled if this is set.
# Supported plugins: abci
plugin = "abci"
```
5. Start the node (from the session with the environment variables set).
6. Observe the log output form the SQS emulator, you should start seeing messages entering the queue.

## Related PRs and Issues

Closes: #211


plugins/abci_indexing/sqs/sqs_client.go
```diff
@@ -0,0 +1,102 @@
...
+)
+
+var queueURLEnvName = "SQS_QUEUE_URL"
+
+type SqsClient struct {
+	queueURL  string
+	sqsClient *sqs.SQS
+}
+
+func (client *SqsClient) sendMessageBatch(batch []*sqs.SendMessageBatchRequestEntry) ([]*sqs.BatchResultErrorEntry, error) {
```
hacheigriega, line 22: `sc` might work better as receiver name according to https://google.github.io/styleguide/go/decisions.html#receiver-names

hacheigriega (commented):
Looks good to me. Adding comments mostly regarding style. Will actually try it out shortly.

hacheigriega (approved):

plugins/abci_indexing/sqs/sqs_client.go
```diff
@@ -0,0 +1,102 @@
...
+	result, err := client.sqsClient.SendMessageBatch(&sqs.SendMessageBatchInput{
+		Entries:  batch,
+		QueueUrl: &client.queueURL,
+	})
+	if err != nil {
+		return nil, err
```
hacheigriega, line 28: The API doc is a little confusing, but I think if `err` is set here, it means that the whole batch has failed. So I guess the first return value should be the whole batch somehow converted into `[]*sqs.BatchResultErrorEntry`. And I don't know how to do that :)

plugins/abci_indexing/sqs/sqs_client.go
```diff
@@ -0,0 +1,102 @@
...
+	}
+
+	return result.Failed, err
```
hacheigriega, line 31: The second return value should be `nil`.

plugins/abci_indexing/sqs/sqs_client.go
hacheigriega: I guess the function can just return the number of failed entries or their IDs instead.

hacheigriega (changes requested):

plugins/abci_indexing/bank/module.go
```diff
@@ -0,0 +1,74 @@
...
+	storetypes "cosmossdk.io/store/types"
+	"github.com/cosmos/cosmos-sdk/codec"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	bankTypes "github.com/cosmos/cosmos-sdk/x/bank/types"
+
+	types "github.com/sedaprotocol/seda-chain/plugins/abci_indexing/types"
+)
+
+func ExtractUpdate(_ codec.Codec, change *storetypes.StoreKVPair) (*types.Message, error) {
+	if keyBytes, found := bytes.CutPrefix(change.Key, bankTypes.SupplyKey); found {
```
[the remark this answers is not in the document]
Thomasvdam: I tried making it more like the simulation decoder, but I'm not 100% sure about the prefix 'pop'. I don't like that the implementation now assumes the prefix is always 1 byte. Is there a way to combine both these methods that I missed?

plugins/abci_indexing/sqs/sqs_client.go
Thomasvdam: Hmm, interesting problem :) I think returning `nil` makes sense since the call site should have access to the `batch` slice and the program can choose how to handle it at that point rather than us having to fabricate some attributes in the `sqs.BatchResultErrorEntry` for all messages. Given that this `err` means no messages were delivered (disregarding ugly network problems).

plugins/abci_indexing/bank/module.go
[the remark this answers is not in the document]
hacheigriega: Yeah that's a good point. Your original implementation might make most sense then, sorry 😬

plugins/abci_indexing/sqs/sqs_client.go
hacheigriega: Yeah that makes sense.

hacheigriega (approved):

PR #216, opened 2024-03-19
Cancel simulation workflows when an update is pushed

Thomasvdam:
## Motivation

This saves some compute when frequently updating PRs.

## Explanation of Changes

I used the same concurrency key that most other workflows are using.

## Testing

Did not test, but since this works for other workflows I'm fairly confident it will work. Famous last words...

## Related PRs and Issues

See the GHA runs for #214, `simulation` was not cancelled while `test` and `interchaintest` were when updates were pushed in quick succession.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -2,11 +2,10 @@
 name: Build
 
 on:
-  push:
-    branches:
-      - main
   pull_request:
-    types: [opened, synchronize, reopened, labeled]
+    push:
+      branches:
+        - main
 
 env:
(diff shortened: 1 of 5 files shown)


hacheigriega (approved):
Thanks for this!! 

A side question: I've been trying to fix the duplicate CI runs in the main of this repo. Do you think we should update the yml files to:
```
on:
  pull_request:
    push:
      branches:
        - main
```
as in the explorer ymls?

hacheigriega:
Done!

mariocao (approved):

Issue #217, opened 2024-03-19
🔧 Indexer plugin logging


Thomasvdam:
## 🔧 Chore

The indexer plugin should be able to log to a file so we can monitor this with tools for errors as well as help with debugging during development. It would be awesome if we can make it log to the host process or stdout but efforts to implement this have so far not gotten us anywhere. Go-plugin makes it look easy, but I worry that on the Cosmos SDK side they're/we're not listening to the logs or doing anything with them.

## Motivation

Right now the indexer plugin does not have any logging capabilities which can be annoying when trying to debug what is happening inside.

Issue #218, opened 2024-03-19
🔧 Indexer plugin error handling


Thomasvdam:
## 🔧 Chore

The error handling in the indexer plugin needs improvement. Currently all errors just cause the plugin to halt. We should at least log information about what happened and have a runbook on how to recover from errors.

## Motivation

Better 'disaster' recovery and information in case something goes wrong.

Blocked by: #217

PR #228, opened 2024-03-26
Plugin logging and error handling

Thomasvdam:
## Motivation

With a logger in place it's a lot easier to debug issues during development and add error monitoring in production. The 'disaster' recovery section explains which actions to take depending on where in the process the error occurred.

## Explanation of Changes

I looked at other Golang projects to see how they handle wrapper around libraries or do error handling so hopefully it's all rather straightforward.

I added the `abort` helper in `plugins/indexing/plugin.go` to make initialisation errors easier to process. For some reason `panic` calls do not (always?) show up in the log output of the node, but if we first log it to stdout they always show. We don't need it in the `ListenX` handlers as the errors returned from these functions are always printed in the node logs.

## Testing

Logging tested by running the chain locally with the following env variables from the repo root:

```sh
make build-plugin-dev && PLUGIN_LOG_FILE="$(pwd)/test.log" PLUGIN_LOG_LEVEL="trace" $BIN start
```

This does assume a chain build is present and the `$BIN` variable points to the executable.

You should see a `test.log` file being created and as blocks get processed entries for both handlers should appear.

Error handling/disaster recovery tested with the following methods:
- Not proving required environment variables when starting the chain.
- Turning off the queue emulator while the chain/plugin were running.
- Manually panicking/returning errors from the `ListenFinalizeBlock` and `ListenCommit` handlers.

## Related PRs and Issues

Closes: #217, #218

### DIFF
--- a/plugins/indexing/log/logger.go
+++ b/plugins/indexing/log/logger.go
@@ -1,0 +1,92 @@
+package logger
+
+import (
+	"fmt"
+	"io"
+	"os"
+
+	"github.com/hashicorp/go-hclog"
+)
+
+var (
+	logFileEnvName  = "PLUGIN_LOG_FILE"
+	logLevelEnvName = "PLUGIN_LOG_LEVEL"
+)
+
+type Logger struct {
+	baseLogger hclog.Logger
+}
+
+func (l *Logger) Trace(msg string, args ...interface{}) {
+	l.baseLogger.Trace(msg, args...)
+}
+
+func (l *Logger) Debug(msg string, args ...interface{}) {
+	l.baseLogger.Debug(msg, args...)
+}
+
+func (l *Logger) Info(msg string, args ...interface{}) {
+	l.baseLogger.Info(msg, args...)
+}
+
+func (l *Logger) Warn(msg string, args ...interface{}) {
+	l.baseLogger.Warn(msg, args...)
+}
+
+func (l *Logger) Error(msg string, args ...interface{}) {
+	l.baseLogger.Error(msg, args...)
+}
+
+// Fatal is equivalent to Error(msg, "error", err) followed by a call to panic(err)
+func (l *Logger) Fatal(msg string, err error) {
+	l.baseLogger.Error(msg, "error", err)
+	// Also printing to stdout so the plugin server error message is more informative
+	fmt.Println(err)
+	panic(err)
+}
+
+func GetLogFile() (*os.File, error) {
+	logFileName, found := os.LookupEnv(logFileEnvName)
+	if !found {
+		err := fmt.Errorf("missing environment variable '%s'", logFileEnvName)
+		return nil, err
+	}
+
+	logFile, err := os.OpenFile(logFileName, os.O_APPEND|os.O_CREATE|os.O_WRONLY, 0o640)
+	if err != nil {
+		return nil, err
+	}
+
+	return logFile, nil
+}
+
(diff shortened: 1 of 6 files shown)


gluax (approved):

plugins/indexing/log/logger.go
```diff
@@ -0,0 +1,84 @@
...
+}
+
+func GetLogFile() (*os.File, error) {
+	logFileName, found := os.LookupEnv(logFileEnvName)
+	if !found {
+		err := fmt.Errorf("missing environment variable '%s'", logFileEnvName)
+		return nil, err
+	}
+
+	logFile, err := os.OpenFile(logFileName, os.O_APPEND|os.O_CREATE|os.O_WRONLY, 0o600)
```
hacheigriega, line 47: Seems a bit too restrictive for a log file? How about 0o640?

plugins/indexing/plugin.go
```diff
@@ -111,3 +132,9 @@ func main() {
 		GRPCServer: plugin.DefaultGRPCServer,
 	})
 }
+
+// Printing the error makes it easier to see what went wrong in the plugin output
+func abort(err error) {
+	fmt.Println(err)
+	panic(err)
+}
```
hacheigriega, line 140: Shouldn't this write the error to the log instead?

plugins/indexing/sqs/sqs_client.go
```diff
@@ -82,21 +82,22 @@ func (sc *SqsClient) PublishToQueue(height int64, data []*types.Message) error {
 	return nil
 }
 
-func NewSqsClient() *SqsClient {
+func NewSqsClient() (*SqsClient, error) {
 	queueURL, found := os.LookupEnv(queueURLEnvName)
 	if !found {
-		panic(fmt.Errorf("missing environment variable '%s'", queueURLEnvName))
+		err := fmt.Errorf("missing environment variable '%s'", queueURLEnvName)
+		return nil, err
```
hacheigriega, line 89: can be just `nil, fmt.Errorf("missing environment variable '%s'", queueURLEnvName)`

plugins/indexing/plugin.go
```diff
@@ -93,14 +102,26 @@ func main() {
...
+		abort(err)
 	}
 	std.RegisterInterfaces(interfaceRegistry)
 	app.ModuleBasics.RegisterInterfaces(interfaceRegistry)
 
+	logFile, err := log.GetLogFile()
+	if err != nil {
+		abort(err)
+	}
+	defer logFile.Close()
```
hacheigriega, line 114: Perhaps this can be moved up so that we can use the logger to write errors from codec registration.

hacheigriega (approved):
LGTM! Left minor comments in case you find them useful.

plugins/indexing/plugin.go
Thomasvdam: Good one, the exception is the case where we're unable to open the log file so there is a little bit of duplication. Also not 100% sure whether this function variable is idiomatic Golang, but to me it made more sense than injecting the logger into the abort function. WDYT?

hacheigriega: Usually a logger has something like `Fatal()` (like here https://cs.opensource.google/go/go/+/master:src/log/log.go;drc=f09db2bb9331f4b31afb867173b1773e6494af27;l=282), but I see that this logger doesn't have that 😅 But I guess we can add a method to `logger.go`?

PR #235, opened 2024-04-02
feat: enable wasm-storage module

hacheigriega:
## Explanation of Changes

Work by @FranklinWaller - I will confirm the proxy contract instantiation and execution implementation before merging.

This PR reenables wasm-storage module to support storing overlay wasms (through governance) and data request wasms. This module also supports proxy contract registry.


Thomasvdam (approved):

hacheigriega:
Converted to draft to focus on mainnet launch for now. Before merging, make sure the following:

- The proxy contract creator should be the sender of the contract (double-check authority flag during transaction creation).
- Merge to a separate branch.
- The contract deployment script runs successfully and is placed in the proper location.

mariocao (changes requested):
Minor thing, but we should reword the "WIP" commit. 😅

hacheigriega:
Ready to be merged, but wanted to double check two things:

- The proxy contract currently only allows the contract creator to execute its functions like `set_staking` and `set_data_requests`. Therefore, I changed the instantiation function to use the admin (the account specified in a flag for `instantiate-and-register-proxy-contract` tx - usually the sender of the tx) as the creator. Before, the creator was set as the gov module account.
- Should we push this to main or should we create a separate branch for Euclid (testnet with DR)?

Cc: @mariocao

deploy_contracts.sh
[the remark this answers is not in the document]
hacheigriega: I'm in favor of leaving it here along with a few other scripts like `local_setup.sh` since they're useful for quickly spinning up local chains.

README.md
```diff
@@ -51,7 +51,8 @@ Download from our Github releases [page](https://github.com/sedaprotocol/seda-ch
 
 ### Build From Source
 
-Please check out [developing](DEVELOPING.md).
+Note that the `main` branch is unstable. For a stable version, `git checkout` a tag that has been released.
```
mariocao, line 54: Just a minor suggestion, but you can leave as it is:
```suggestion
Note that the `main` branch is under development and might not always be stable. For a stable version, consider checking out a released tag using `git checkout`.
```

mariocao (approved):

README.md
hacheigriega: perfect. updated

PR #238, opened 2024-04-04
doc: update Lavender.Five snapshot link

hacheigriega:
Updating Lavender.Five's snapshot link.
We will rely on this until we provide snapshots ourselves.

### DIFF
--- a/README.md
+++ b/README.md
@@ -85,27 +85,28 @@ When connecting externally, choose a trustworthy node operator. Unscrupulous ope
 ### Running the Node Yourself
 
 ```
-<!-- rename the downloaded binary to a simpler name -->
+# Rename the downloaded binary to a simpler name.
 mv sedad-${ARCH} sedad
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -85,27 +85,28 @@ When connecting externally, choose a trustworthy node operator. Unscrupulous ope
 ### Running the Node Yourself
 
 ```
-<!-- rename the downloaded binary to a simpler name -->
+# Rename the downloaded binary to a simpler name.
```
mariocao, line 88: Maybe add something like this?

```suggestion
# Rename the downloaded binary to a simpler name.
export ARCH=$(uname -m)
```

mariocao: in linux it returns `x86_64`

mariocao: Maybe something like this:
```
export ARCH=$(arch | sed s/aarch64/arm64/ | sed s/x86_64/amd64/)
```

mariocao (approved):

PR #241, opened 2024-04-09
Feat: Add Default Time For vesting time in CSV Script

gluax:
## Motivation

So if in the CSV file they do not provide the start time wich aligns more the traditional vesting time command.

## Explanation of Changes

Script now:
- checks the CSV and genesis files exist.
- stores the Genesis time before reading the CSV.
- Checks if the vesting amount is set; if the start time is not set, it is to the genesis time. 

## Testing

Added a new line to the example csv to test this with.

## Related PRs and Issues

N/A


hacheigriega (approved):
Perfect. Thanks Jon

Issue #247, opened 2024-04-17
🐛 Large messages are not published


Thomasvdam:
## 🐛 Bug Report

SQS messages that are too large can not be published, halting the indexing process. See https://sedaprotocol.github.io/test-explorer/devnet/tx/9EAE8609BAB8553952D592211E1B93ACF4DF613E7A9856848B00207C0E7A74E7 for an example of a TX that failed to index.

## Steps to Reproduce

Submit a transaction with a large message body, contract uploads are an easy one:

```json
{
  "tx": {
    "body": {
      "messages": [
        {
          "@type": "/cosmos.group.v1.MsgSubmitProposal",
          "group_policy_address": "seda1rkml5878l2daw3a7xvg48wqecnh9u9dn2dtl8g57rsctq5pnc00sue22w3",
          "proposers": [
            "seda1j69tj7k4kl998gj833fdmy8fm3umkxa7ytvnwh"
          ],
          "metadata": "",
          "messages": [
            {
              "@type": "/cosmwasm.wasm.v1.MsgStoreAndInstantiateContract",
              "authority": "seda1rkml5878l2daw3a7xvg48wqecnh9u9dn2dtl8g57rsctq5pnc00sue22w3",
              "wasm_byte_code": "OMITTED_FOR_BREVITY",
                "instantiate_permission": null,
                "unpin_code": false,
                "admin": "seda1rkml5878l2daw3a7xvg48wqecnh9u9dn2dtl8g57rsctq5pnc00sue22w3",
                "label": "Wormhole Core Contract (Testnet)",
                "msg": {
                  "gov_chain": 1,
                  "gov_address": "AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAQ=",
                  "guardian_set_expirity": 86400,
                  "initial_guardian_set": {
                    "addresses": [
                      {
                        "bytes": "E5R71IsY5T/a7ud/NHM5Gscnxjg="
                      }
                    ],
                    "expiration_time": 0
                  },
                  "chain_id": 4005,
                  "fee_denom": "aseda"
                },
                "funds": [],
                "source": "",
                "builder": "",
                "code_hash": null
              }
            ],
            "exec": "EXEC_UNSPECIFIED",
            "title": "Wormhole Core (Testnet)",
            "summary": "Deploy the Wormhole Core contract configured for Ethereum Testnet (Sepolia)"
          }
        ],
        "memo": "",
        "timeout_height": "0",
[...]

PR #248, opened 2024-04-18
Publish large message payloads to S3

Thomasvdam:
## Motivation

The 'official' recommendation from Amazon is to offload larger message bodies to S3 and publish a message with a reference to the uploaded file, which is why we also went for this approach.

## Explanation of Changes

To have some leeway we upload to S3 at ~100KB rather than the exact limit of SQS. This means we should realistically always be able to publish the upload message body with the 'normal' message attributes.

Also adds a bunch of extra logs to make debugging things easier when it comes to SQS related issues.

## Testing

You'll need to set up https://github.com/adobe/S3Mock on your local machine (the docker image is super easy to spin up). Once you have it running make sure to create a bucket before doing anything else: `curl --request PUT "http://localhost:9444/indexer-localnet-large-messages/"`. You should also have the SQS emulator running. If you're not using the `seda-explorer` repository devcontainer docker-compose file as a reference please double check all the ports used in the commands.

>[!CAUTION]
> To make life easier add the following 2 lines to `./scripts/local_setup.sh` below the "# configure sedad" comment on line 32 
> ```sh
> $BIN config set app streaming.abci.keys '["*"]'
> $BIN config set app streaming.abci.plugin '"abci"'
> ```

Now build the plugin in dev mode and start the chain with all the expected environment variables:

```sh
make build-plugin-dev && COSMOS_SDK_ABCI=PATH_TO_REPOSITORY_ROOT/seda-chain/build/plugin SQS_QUEUE_URL=http://localhost/4100/local-updates.fifo SQS_ENDPOINT=http://localhost:4100 S3_ENDPOINT=http://localhost:9444 S3_LARGE_MSG_BUCKET_NAME="indexer-localnet-large-messages" PLUGIN_LOG_FILE=./test.log PLUGIN_LOG_LEVEL=trace ./scripts/local_setup.sh
```

Once the chain is running and past block 1 you can submit a code upload transaction. I used the wormhole core contract, but most contracts should be fine:

```sh
$BIN tx wasm store PATH_TO_LARGE_WASM --from acc1 --keyring-backend test --gas 300000000000 -y
```

Once the block with this TX is committed you should see that a file was uploaded to the S3 mock under the key `"tx-h${BLOCK_HEIGHT}-i${MSG_INDEX}.json"` with the JSON payload as the message body, and on the queue you should see a message with the following payload:

```json
{
  "type": "large-message",
  "data": { "key": "tx-h${BLOCK_HEIGHT}-i${MESSAGE_INDEX}.json", "ETag": "\"74c6403e1d2713ff795a958be03e38ff\"" }
}
```

## Related PRs and Issues

Closes: #247


plugins/indexing/aws/sqs_client.go
```diff
@@ -0,0 +1,69 @@
+package pluginaws
```
hacheigriega, line 1: Hm I thought package name always had to match the directory name (`aws` in this case) in golang..

plugins/indexing/aws/sqs_client.go
```diff
@@ -0,0 +1,69 @@
...
+
+	sizedBatchEntries, err := sc.createSizedBatchEntries(data)
+	if err != nil {
+		return err
+	}
+
+	currentRequestSize := 0
+	entries := make([]*sqs.SendMessageBatchRequestEntry, 0, MaxSQSBatchSize)
+	for _, sizedBatchEntry := range sizedBatchEntries {
+		nextSize := currentRequestSize + sizedBatchEntry.size
```
hacheigriega, line 27: It seems this variable is unnecessary - just a minor suggestion though.

hacheigriega (approved):

plugins/indexing/aws/sqs_client.go
hacheigriega: I checked that it's just a convention. Up to you I guess ultimately..

hacheigriega: So true what they say about variable naming!

hacheigriega:
Not related to this PR, but I just found that the package name of `plugins/indexing/auth` is `bank`.

plugins/indexing/aws/sqs_client.go
hacheigriega: Upon checking Geth I believe we should follow the convention and name the package `aws` even if it sounds generic.

plugins/indexing/aws/sqs_client.go
Thomasvdam: You mean you would leave out the variable and 'inline' the addition in the check? Or is there a cleaner way to do it?

hacheigriega: Yeah inline. Or assign `nextSize` to `currentRequestSize` directly at the end, I guess. Minor suggestion.

Thomasvdam: Sorry if I’m being dense but I’m not sure if I understand correctly. I use nextSize to determine whether adding the current message will make the batch go over the size limit or not so the only place I would need it is in the check jut below. My reasoning to extract from the statement was for legibility, but if this is not good practice I can just move it into the check. :)

hacheigriega: My bad for confusing wording.. I think we can just change line 39 to:
```
currentRequestSize = nextSize
```

mariocao (approved):

Issue #42, opened 2023-09-06
✨ Pruning strategies for Wasm binaries


hacheigriega:
## ✨ Feature

`x/wasm-storage` module needs a queue structure to process pruning. It still has to be determined whether the queue would be keyed by expiration timestamp or expiration block number. We could provide a few pruning strategy options and allow the user to pick one using a transaction flag.

DeshErBojhaa:
Feature List
1. Store WASM Binary [Done].
2. Prune WASM binaries when the lifespan is over. 

- Add user address as a property in `types.Wasm`
- Fee for storing WASM = `size` * `duration` * `UnitPrice`. Where `UnitPrice` is a module parameter.
- Every time a WASM is used, its life span is extended by `PersisPeriod`. Where `PersisPeriod` is a module parameter.

DeshErBojhaa:
## PRs
Refactor existing code https://github.com/sedaprotocol/seda-chain/pull/263

PR #266, opened 2024-05-12
feat: added exp block for data request wasm

DeshErBojhaa:
## Motivation

This PR implements the feature described [in this PR](https://github.com/sedaprotocol/seda-chain/issues/42)

## Explanation of Changes

- Added a prune height in `Wasm` type. Only DR wasms have a prune block height. Overlay Wasms are never pruned.

## Testing

Tested using keeper. 
Future plan: Testing availing the ABCI call `FinalizeBlock`

## Related PRs and Issues

[Pruning strategies for Data Request Wasm binaries](https://github.com/sedaprotocol/seda-chain/issues/42)


x/wasm-storage/keeper/msg_server.go
```diff
@@ -56,19 +59,38 @@ func NewMsgServerImpl(keeper Keeper) types.MsgServer {
...
+	}
 
-	unzipped, err := unzipWasm(msg.Wasm)
+	unzipped, err := unzipWasm(msg.Wasm, params.MaxWasmSize)
 	if err != nil {
 		return nil, err
 	}
-	wasm := types.NewWasm(unzipped, msg.WasmType, ctx.BlockTime())
+
+	wasm := types.NewWasm(unzipped, msg.WasmType, ctx.BlockTime(), ctx.BlockHeight(), params.WasmTTL)
```
hacheigriega, line 81: Perhaps `NewWasm` should return a pointer so that we don't have to dereference it?

x/wasm-storage/types/wasm.go
```diff
@@ -15,29 +15,30 @@ const (
...
 	return &Wasm{
-		Hash:     hash,
-		Bytecode: bytecode,
-		WasmType: wasmType,
-		AddedAt:  addedAt,
+		Hash:        hash,
+		Bytecode:    bytecode,
+		WasmType:    wasmType,
+		AddedAt:     addedAt,
+		PruneHeight: curBlock + ttl,
```
hacheigriega, line 41: `ExpirationHeight`?

x/wasm-storage/keeper/msg_server.go
```diff
@@ -56,19 +59,38 @@ func NewMsgServerImpl(keeper Keeper) types.MsgServer {
...
 	}
 
-	if err := m.DataRequestWasm.Set(ctx, WasmKey(*wasm), *wasm); err != nil {
+	wasmKey := WasmKey(*wasm)
+	if err := m.DataRequestWasm.Set(ctx, wasmKey, *wasm); err != nil {
+		return nil, err
+	}
+
+	// Track Exp.
+	expKey := collections.Join(wasm.PruneHeight, wasmKey)
```
hacheigriega, line 92: No need for this variable + Fix comment to "Track expiration."

hacheigriega (approved):

x/wasm-storage/types/wasm.go
hacheigriega: Related to previous comment, but I think it's clearer to set this field outside of this constructor function.

x/wasm-storage/keeper/msg_server.go
```diff
@@ -96,16 +118,20 @@ func (m msgServer) StoreOverlayWasm(goCtx context.Context, msg *types.MsgStoreOv
...
 		return nil, fmt.Errorf("invalid authority %s", msg.Sender)
 	}
 
-	unzipped, err := unzipWasm(msg.Wasm)
+	unzipped, err := unzipWasm(msg.Wasm, params.MaxWasmSize)
 	if err != nil {
 		return nil, err
 	}
-	wasm := types.NewWasm(unzipped, msg.WasmType, ctx.BlockTime())
+	wasm := types.NewWasm(unzipped, msg.WasmType, ctx.BlockTime(), ctx.BlockHeight(), math.MaxInt64) // Forever
```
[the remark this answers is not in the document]
DeshErBojhaa: It will surely overflow. Great catch!!
I will use -1 as forever.

x/wasm-storage/keeper/keeper.go
```diff
@@ -19,11 +19,13 @@ var (
 	DataRequestPrefix = collections.NewPrefix(0)
 	// OverlayPrefix defines prefix to store Overlay Wasm binaries.
 	OverlayPrefix = collections.NewPrefix(1)
+	// WasmExpPrefix defines prefix to track wasm expiration.
+	WasmExpPrefix = collections.NewPrefix(2)
```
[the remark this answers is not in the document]
DeshErBojhaa: Keys are moved to `types/keys.go`

About the naming, This data structure does not maintain any order. The work queue means strong order of FIFO. So I am keeping the name as is.

x/wasm-storage/types/tx.go
```diff
@@ -7,59 +7,59 @@ import (
...
+func (msg *MsgStoreOverlayWasm) ValidateBasic() error {
 	if _, err := sdk.AccAddressFromBech32(msg.Sender); err != nil {
 		return err
 	}
 	if msg.WasmType != WasmTypeDataRequestExecutor && msg.WasmType != WasmTypeRelayer {
 		return fmt.Errorf("overlay Wasm type must be data-request-executor or relayer")
 	}
-	if err := validateWasmCode(msg.Wasm); err != nil {
-		return fmt.Errorf("invalid request: code bytes %s", err.Error())
+	if err := validateWasmLen(msg.Wasm); err != nil {
```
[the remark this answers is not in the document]
DeshErBojhaa: Addressed here: [04f76f0](https://github.com/sedaprotocol/seda-chain/pull/266/commits/04f76f0a0e0b2347c048194e7b68f496b41eedd7)

x/wasm-storage/keeper/msg_server.go
DeshErBojhaa: Addressed here: [a1f2999](https://github.com/sedaprotocol/seda-chain/pull/266/commits/a1f29999f6d95815f78110a6d6c910f6f3359614)

x/wasm-storage/types/params.go
```diff
@@ -28,7 +28,7 @@ func (p *Params) ValidateBasic() error {
 	return validateMaxWasmSize(p.MaxWasmSize)
 }
 
-func validateMaxWasmSize(i uint64) error {
+func validateMaxWasmSize(i int64) error {
 	if i == 0 {
```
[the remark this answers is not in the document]
DeshErBojhaa: Here: [03d50ff](https://github.com/sedaprotocol/seda-chain/pull/266/commits/03d50ff1e022660b4ecdb43672cc2c1f7f1045a9)

x/wasm-storage/keeper/msg_server.go
DeshErBojhaa: [03d50ff](https://github.com/sedaprotocol/seda-chain/pull/266/commits/03d50ff1e022660b4ecdb43672cc2c1f7f1045a9)

mariocao (commented):
Why changing from unsigned integers to signed integers?
For example, in the `prune_height`, `max_wasm_size` and `wasm_ttl`.

mariocao:
> Why changing from unsigned integers to signed integers? For example, in the `prune_height`, `max_wasm_size` and `wasm_ttl`.

After a quick check, I saw that in other Cosmos SDK native modules they also use `int64` for block height. :/

x/wasm-storage/keeper/keeper.go
```diff
@@ -158,6 +148,43 @@ func (k Keeper) GetAllWasms(ctx sdk.Context) []types.Wasm {
...
+//	wasmKeys, err := keeper.WasmKeyByExpBlock(ctx, expHeight)
+//	if err != nil {
+//	    return err
+//	}
+//	for _, key := range wasmKeys {
+//	    fmt.Println(string(key))
+//	}
+//
+// Note: The returned keys do not maintain an order.
+func (k Keeper) WasmKeyByExpBlock(ctx sdk.Context, expBlock int64) ([][]byte, error) {
```
mariocao, line 169: After taking a look to other keepers, in our repo and in Cosmos SDK modules, I would call this function something like `GetExpiredWasmKeys`.

x/wasm-storage/keeper/keeper.go
```diff
@@ -14,25 +14,13 @@ import (
...
-	OverlayPrefix = collections.NewPrefix(1)
-	// ProxyContractRegistryPrefix defines prefix to store address of
-	// Proxy Contract.
-	ProxyContractRegistryPrefix = collections.NewPrefix(2)
-	// ParamsPrefix defines prefix to store parameters of wasm-storage module.
-	ParamsPrefix = collections.NewPrefix(3)
-)
-
 // WasmKey takes a wasm as parameter and returns the key.
 func WasmKey(w types.Wasm) []byte {
```
mariocao, line 18: Same here... `GetWasmKey`

x/wasm-storage/keeper/keeper.go
```diff
@@ -46,6 +34,7 @@ type Keeper struct {
 	Schema                collections.Schema
 	DataRequestWasm       collections.Map[[]byte, types.Wasm]
 	OverlayWasm           collections.Map[[]byte, types.Wasm]
+	WasmExp               collections.KeySet[collections.Pair[int64, []byte]]
```
hacheigriega, line 37: `WasmExpirations` ?

hacheigriega (changes requested):
Great work!! I just thought some naming could be improved.

hacheigriega:
> And one last doubt I have is whether `KeySet` is better than `Map`. If we use a map with the array of wasm hashes as values, we can avoid iterating based on prefixes, right? I'm thinking that might be more efficient.

I see that Cosmos's recommended implementation uses `Map` structure to process unbonding. https://github.com/testinginprod/cosmos-sdk/pull/22/files#diff-bd9428f4e6e0c1385c6122f03a090b2ed52aa6bae9ea4a896b61917e5a858796R63

x/wasm-storage/keeper/keeper.go
DeshErBojhaa: This is not a queue. On gurantee of order.

proto/sedachain/wasm_storage/v1/wasm_storage.proto
```diff
@@ -15,7 +15,7 @@ message Wasm {
       [ (gogoproto.stdtime) = true, (gogoproto.nullable) = false ];
   // prune_height represents the block height till which the Wasm will
   // stay on the chain. At height > prune_height, the wasm should be pruned.
-  uint64 prune_height = 5;
+  int64 prune_height = 5;
```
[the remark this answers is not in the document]
DeshErBojhaa: here: [9cd4dcd](https://github.com/sedaprotocol/seda-chain/pull/266/commits/9cd4dcdfb9578b193a3adce22a0095e8ab17c507)

x/wasm-storage/keeper/keeper.go
DeshErBojhaa: in this commit: [9cd4dcd](https://github.com/sedaprotocol/seda-chain/pull/266/commits/9cd4dcdfb9578b193a3adce22a0095e8ab17c507)

x/wasm-storage/keeper/keeper.go
DeshErBojhaa: in this commit: [9cd4dcd](https://github.com/sedaprotocol/seda-chain/pull/266/commits/9cd4dcdfb9578b193a3adce22a0095e8ab17c507)

jim380:
> Great work!! I just thought some naming could be improved. And one last doubt I have is whether `KeySet` is better than `Map`. If we use a map with the array of wasm hashes as values, we can avoid iterating based on prefixes, right? I'm thinking that might be more efficient.

I actually think that with the way `KeySet` is currently implemented in Collections the difference vs. `Map` would be rather small. `KeySet` is precisely a `Map` with no value, but internally each `KeySet` does have a `[]byte{}`, which does allocate memory unlike `struct{}`.

I tend to lean towards Map in most cases.

DeshErBojhaa:
Why collections.Map is not used.

Current implementation of Map can not arbitrary take any key, value combination.
Any data to be used as key need to satisfy `codec.KeyCodec` interface.
And for a value it need to satisfy `codec.ValueCodec`

I we were to use map in this case, we have to either:
1. Introduce a new proto to hold the list of wasms.
2. Implement `codec.ValueCodec` on a [][]bytes.

Which is over kill for this straightforward task.

jim380:
Also, even if you used a `collections.Map`, you'd still have to use `collections.NewPrefixedPairRange` to iterate over the composite keys. `KeySet` is a `Map` at the end of the day 😄

hacheigriega (approved):

Issue #258, opened 2024-04-30
🔧 Add message type filter to plugin


Thomasvdam:
## 🔧 Chore

Right now the plugin will publish all messages that are created from the store updates. We should add a mechanism to allowlist certain message types.

## Motivation

Allow a node to catch up without having to broadcast messages that have already been broadcast previously and are already indexer.

An allowlist makes more sense as we'll usually only introduce a few new messages which require a re-index of the previous history.

## Implementation

- [ ] Design method for passing the message types that are allowed.
- [ ] Make default so that it passes all messages.
- [ ] Add filter at the correct level (probably SQS, but maybe earlier in the process makes sense?)

PR #268, opened 2024-05-16
Add message filtering capability

Thomasvdam:
## Motivation

Make indexing previously indexed state easier by skipping messages that have already been indexed.

## Explanation of Changes

Optionally allow configuring a list of message types that are allowed to be published on the queue. Went for an allowlist as we'll generally use it when we want to reindex historical state and only publish updates for the new messages.

For the same reason I went for a simple slice of strings as the data structure; the list should never be long enough to matter in terms of performance.

## Testing

Ran a local chain with the messages allowlist attribute set to `account,account-balance`

Logs:
```log
2024-05-15T18:00:34.456+0800 [TRACE] publishing to queue: size=4
2024-05-15T18:00:34.456+0800 [TRACE] skipping message: type=supply
2024-05-15T18:00:34.456+0800 [TRACE] sending batch: size=3
```

And on the queue side we only received the account and account balance updates.

## Related PRs and Issues

Closes: #258

### DIFF
--- a/plugins/indexing/README.md
+++ b/plugins/indexing/README.md
@@ -17,6 +17,9 @@ export S3_LARGE_MSG_BUCKET_NAME=""
 export PLUGIN_LOG_FILE=PATH_TO_DESIRED_LOG_FILE
 # Optionally you can also specify the log level, one of "trace", "debug", "info", "warn", "error"
 export PLUGIN_LOG_LEVEL="WARN"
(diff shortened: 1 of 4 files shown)


hacheigriega (approved):
I suppose it would be more efficient to use a map for `SqsClient.allowedMessages` but I see your point that the list would be small anyways.

PR #271, opened 2024-05-17
refactor: refactor wasm storage module collections

hacheigriega:
Fixing some issues introduced during the migration to collections. 

- Fix the issue where we add wasm type prefix unnecessarily.
- Update field names so they are clearer.
- Remove unnecessary functions.


Thomasvdam:
This also includes a bunch of the tally VM stuff, is that intended?

hacheigriega:
> This also includes a bunch of the tally VM stuff, is that intended?

removed the tally vm import, which is not relevant here!

DeshErBojhaa (approved):
🎉 Great work!!

x/wasm-storage/keeper/test_utils/debug.wasm
Thomasvdam, line 1: Is this not also part of the tally?

PR #280, opened 2024-06-10
feat: Filtering implementation with support for `None` filter type

DeshErBojhaa:
## Motivation

Filtering fro data request wasm reveals.

## Explanation of Changes

Right now only NONE and MODE are implemented.
## Testing

Unit test included.


x/wasm-storage/keeper/abci.go
```diff
@@ -109,3 +126,18 @@ func (k Keeper) ExecuteTally(ctx sdk.Context) error {
 
 	return nil
 }
+
+func tallyVMArg(reveals map[string]any, outliers []bool) ([]string, error) {
+	argBytes, err := rlp.EncodeToBytes(tallyArg{
```
FranklinWaller, line 131: Should be JSON. From the Tally VM notion:

* REVEALS: An JSON encoded array of reveals coming from the contract. The shape of this JSON is out of scope of this document. Refer to the contract for an accurate format
* OUTLIARS: An JSON encoded array with the outliars that are calculated by the filter mechanism. Ex: [0, 0, 1, 1, 1, 1] Where the 1 means its considered correct and 0 means it was out of consensus.

drfilters/filter.go
```diff
@@ -0,0 +1,45 @@
...
+)
+
+const (
+	None         = 0
+	Mode         = 1
+	StdDeviation = 2
+)
+
+type (
+	tallyProp struct {
```
FranklinWaller, line 16: It's better to have a struct for each filter type instead of doing optional. And have a prestep where you decode the rlp and switch case on the mode to get the correct struct

mariocao (changes requested):
Apart from Franklin's comments, I want to add that you have duplicated commits in the PR.

drfilters/filter.go
```diff
@@ -0,0 +1,45 @@
+package drfilters
```
hacheigriega, line 1: No need to create a package in my opinion. We can just add `filter.go` under `x/wasm-storage/keeper`.

drfilters/filter.go
```diff
@@ -0,0 +1,45 @@
...
+	"errors"
+
+	"github.com/ethereum/go-ethereum/rlp"
+)
+
+const (
+	None         = 0
+	Mode         = 1
+	StdDeviation = 2
+)
```
hacheigriega, line 13:
```
type FilterType int32

const (
    FILTER_TYPE_NONE FilterType = 0
    FILTER_TYPE_MODE FilterType = 1
    FILTER_TYPE_STD_DEV FilterType = 2
)
```

drfilters/filter.go
DeshErBojhaa: The approach you mentioned is difficult to implement. It's not possible to know the filtering mode, without decoding to the struct first.

x/wasm-storage/keeper/abci.go
DeshErBojhaa: Here: [b12fb80](https://github.com/sedaprotocol/seda-chain/pull/280/commits/b12fb80e4341ab1eec6c1e5a0921759ffff97dfa)

drfilters/filter.go
DeshErBojhaa: Here: [0fe4111](https://github.com/sedaprotocol/seda-chain/pull/280/commits/0fe411151ee1fa892813b31547dce3c1d576de40)

drfilters/filter.go
DeshErBojhaa: good point. here: [d6e8806](https://github.com/sedaprotocol/seda-chain/pull/280/commits/d6e88060c93ec98e2060560ce5736feba66a0f17)

x/wasm-storage/keeper/abci.go
```diff
@@ -109,3 +139,19 @@ func (k Keeper) ExecuteTally(ctx sdk.Context) error {
...
+		Outliers:       outliers,
+	})
+	if err != nil {
+		return nil, err
+	}
+	arg := make([]string, 0, len(argBytes))
+	for _, b := range argBytes {
+		arg = append(arg, fmt.Sprintf("%c", b))
+	}
+	return arg, err
```
FranklinWaller, line 156: Maybe i am wrong (To be honest i find Go a lil harder to read what exactly is going on here) but this doesn't seem like it complies with the format:

```args: ["TALLY_INPUT_ARGS", "REVEALS", "OUTLIARS"]``` or is it? if so you can ignore my comment :)

File: x/wasm-storage/keeper/filter.go [hunk 6]
```diff
@@ -0,0 +1,147 @@
...
+
+import (
+	"errors"
+
+	"github.com/ethereum/go-ethereum/rlp"
+	"github.com/tidwall/gjson"
+)
+
+const (
+	None         = iota
```
FranklinWaller, line 11: Is iota not the same as 0?

x/wasm-storage/keeper/filter.go
```diff
@@ -0,0 +1,147 @@
...
+			vals = append(vals, res.Float())
+		}
+		outliers, consensus = calculate(vals)
+	default:
+		vals := make([]string, 0, len(reveals))
+		for _, res := range results {
+			vals = append(vals, res.String())
+		}
+		outliers, consensus = calculate(vals)
+	}
```
FranklinWaller, line 80: We can't do in general `res.String()` instead of handling each case? Don't go have some interface where you can always convert everything to String?

drfilters/filter.go
FranklinWaller: In rust we decode by hand, couldn't you get the first index of RLP and then decode into a struct?
The thing is RLP is indexed based and if we have different properties in the future then we have to zero out optional properties on a protocol level which seems weird to do just for Go, while other languages handle this fine.

For example in rust:
```
fn extract_bytes(tokens: &[Token], index: usize, name: &str) -> Result<Vec<u8>, RelayerError> {
    extract_token(tokens, index, name, |t| match t {
        Token::FixedBytes(bytes) => Some(bytes),
        other => other.into_bytes(),
    })
}
```

x/wasm-storage/keeper/filter.go
DeshErBojhaa: Yes. 0 then auto increment each line

x/wasm-storage/keeper/filter.go
DeshErBojhaa: here: [92b66cf](https://github.com/sedaprotocol/seda-chain/pull/280/commits/92b66cfd5aacfb22659422df22d0eb81b458e261)

x/wasm-storage/keeper/abci.go
DeshErBojhaa: here: [d6f112b](https://github.com/sedaprotocol/seda-chain/pull/280/commits/d6f112bf0fe59b8de70094875438195d7233d1e5)

drfilters/filter.go
DeshErBojhaa: here: [c015652](https://github.com/sedaprotocol/seda-chain/pull/280/commits/c015652b646e34e940d39e677ac65988057a6a61)

x/wasm-storage/keeper/abci.go
```diff
@@ -4,34 +4,40 @@ import (
...
+		ID       []byte `json:"id"`
+		// Memo              []interface{}          `json:"memo"`
+		// PaybackAddress    []interface{}          `json:"payback_address"`
+		ReplicationFactor int                   `json:"replication_factor"`
+		Reveals           map[string]RevealBody `json:"reveals"`
+		// SedaPayload       []interface{}          `json:"seda_payload"`
+		TallyBinaryID []byte `json:"tally_binary_id"`
+		TallyInputs   []byte `json:"tally_inputs"`
+		Version       string `json:"version"`
+	}
```
mariocao, line 30: Why are some fields commented out?

x/wasm-storage/keeper/abci.go
```diff
@@ -88,16 +94,34 @@ func (k Keeper) ExecuteTally(ctx sdk.Context) error {
 
 	// 3. Loop through the list to execute tally.
 	// TODO: is it ok to use a map?
```
mariocao, line 96: Does this TODO still make sense?

x/wasm-storage/keeper/filter.go
```diff
@@ -0,0 +1,121 @@
...
+		Algo     uint
+		JSONPath string
+	}
+
+	//stdFilter struct {
+	//	Algo       uint
+	//	JSONPath   string
+	//	MaxSigma   uint64
+	//	NumberType uint8
+	//}
```
mariocao, line 27: Remove unused code?

mariocao (commented):
I missed some testing for the mode filtering.

x/wasm-storage/keeper/abci.go
DeshErBojhaa: @hacheigriega cc.

x/wasm-storage/keeper/abci.go
DeshErBojhaa: @hacheigriega cc.

x/wasm-storage/keeper/filter.go
```diff
@@ -0,0 +1,121 @@
...
+
+	"github.com/ethereum/go-ethereum/rlp"
+	"github.com/tidwall/gjson"
+)
+
+const (
+	None = iota
+	Mode
+	StdDeviation
+)
```
hacheigriega, line 14: Let's create a type `FilterType`?
```
type FilterType byte

const (
	None FilterType = iota
	Mode
	StdDev
)
```

x/wasm-storage/keeper/abci.go
hacheigriega: the base branch has been updated

x/wasm-storage/keeper/abci.go
hacheigriega: Let's understand the discussion around determinism & maps to make sure we're safe.
Ex: https://github.com/cosmos/cosmos-sdk/issues/13039

x/wasm-storage/keeper/filter.go
DeshErBojhaa: Do we have any use case of the new type Filtertype as of now? If not, we don't have to complicate by introducing another type.

x/wasm-storage/keeper/abci.go
DeshErBojhaa: Since the code under the loop can cause error(s). Let's have the keys sorted. So that even if we error out, we ensure all the nodes errored on the same map item.

hacheigriega: Or just modify the contract code to return a vector, not a hashmap :) That would be a better solution, right? Cc @FranklinWaller @gluax

x/wasm-storage/keeper/filter_test.go
```diff
@@ -0,0 +1,42 @@
...
+			wantErr: nil,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			got, _, err := keeper.Outliers(tt.tallyInput, tt.reveals)
+			if tt.wantErr != nil {
+				require.ErrorIs(t, err, tt.wantErr)
+				return
+			}
```
hacheigriega, line 38: This only checks for an error when we expect an error

hacheigriega: And I think there is in fact an error here.

x/wasm-storage/keeper/filter.go
hacheigriega: I see what you are saying. I converted it to byte so that we can compare the first byte of tally inputs directly to these.

x/wasm-storage/keeper/filter.go
```diff
@@ -74,19 +79,29 @@ func calculate[T comparable](reveals []T) ([]bool, bool) {
...
-	var filter filterProp
-	if err := rlp.DecodeBytes(filterInput, &filter); err != nil {
+	var rlpAsList []any
+	if err := rlp.DecodeBytes(filterInput, &rlpAsList); err != nil {
 		return nil, false, err
 	}
+	filteringAlgo, ok := rlpAsList[0].([]uint)
+	if !ok || len(filteringAlgo) != 1 {
+		return nil, false, errors.New("can not RLP decode algo type from filter input")
+	}
```
hacheigriega: Was confused about these lines. Why did you use RLP decoding here? I changed the code so that it just looks at the first byte. Let me know what you think.

x/wasm-storage/keeper/filter.go
```diff
@@ -0,0 +1,120 @@
...
+	switch tallyInputs[0] {
+	case filterNone:
+		outliers := make([]bool, len(reveals))
+		for i := range outliers {
+			outliers[i] = false
+		}
+		return outliers, true, nil
+
+	case filterMode:
+		// TODO: Reactivate mode filter
```
hacheigriega, line 93: Removed this for now - let's address this in a separate PR cc @DeshErBojhaa

hacheigriega (approved):

PR #285, opened 2024-06-20
feat: Tally execution

### LINKED ISSUE #125: ✨ x/tally: Execution phase
## ✨ Feature

The last phase of Tally phase involves fetching and executing the tally binary corresponding to a given data request binary, and then reporting the result back to the contract. 

Currently, the VM execution fails with the following error:
```
{[] [abort: Key does not exist in ~lib/map.ts(105:17)
] [] { 255}}
{[] [abort: Key does not exist in ~lib/map.ts(105:17)
] [] { 255}}
```

hacheigriega:
# Explanation

This PR implements execution of tally and posting the execution results to the SEDA contract.

An integration test has been added to test the flow from a mocked fetching of a tally list from the contract up to the point where the execution results are posted to the contract. This sudo call to post the execution results fails as expected since the contract does not actually contain request data. 

- [x] Tally VM execution
- [x] Tally VM tests
- [x] Post tally execution results
- [ ] Improve error handling in end blocker (should may be addressed in a separate PR)

### Leftover Issues

- [ ] `PostDataResults`'s message structure does not seem to support submission of multiple results at once. (contract needs update - perhaps not a priority)
- [ ] Use limits when fetching tally list from contract.
- [ ] Agree on `VMResult` data type format.
- [x] Add extensive logging to facilitate debugging.

Closes #125


x/wasm-storage/keeper/filter.go
```diff
@@ -14,16 +14,16 @@ const (
...
-func ApplyFilter(tallyInputs []byte, reveals []RevealBody) ([]bool, bool, error) {
+func ApplyFilter(tallyInputs []byte, reveals []RevealBody) ([]int, bool, error) {
 	if len(tallyInputs) < 1 {
 		return nil, false, errors.New("tally inputs should be at least 1 byte")
 	}
 
 	switch tallyInputs[0] {
 	case filterNone:
-		outliers := make([]bool, len(reveals))
+		outliers := make([]int, len(reveals))
```
DeshErBojhaa, line 24: why change to int?

x/wasm-storage/keeper/filter.go
```diff
@@ -14,16 +14,16 @@ const (
...
 		for i := range outliers {
-			outliers[i] = false
+			outliers[i] = 0
```
DeshErBojhaa, line 26: no need to assign value. default is 0

DeshErBojhaa (changes requested):
I need to sync with @hacheigriega for further review.

x/wasm-storage/keeper/filter.go
hacheigriega: The only reason I made this change was because the VM expects the outlier list in JSON array of numbers

DeshErBojhaa (approved):
🎉

Issue #286, opened 2024-06-21
✨ Decode filtering RLP


hacheigriega:
## ✨ Feature

The current tally implementation simply checks the first byte of `tally_inputs` to determine which filter type to run. Since we have decided to add a new field to `Request` to encode filtering parameters in RLP, we need to update the implementation accordingly.

DeshErBojhaa:
We can introduce a new field in the `Request` 

```go
type Request struct {
	DrBinaryID        string                `json:"dr_binary_id"`
	DrInputs          string                `json:"dr_inputs"`
	GasLimit          string                `json:"gas_limit"`
	GasPrice          string                `json:"gas_price"`
	Height            uint64                `json:"height"`
	ID                string                `json:"id"`
	Memo              string                `json:"memo"`
	PaybackAddress    string                `json:"payback_address"`
	ReplicationFactor int64                 `json:"replication_factor"`
	Reveals           map[string]RevealBody `json:"reveals"`
	SedaPayload       string                `json:"seda_payload"`
	TallyBinaryID     string                `json:"tally_binary_id"`
	TallyInputs       string                `json:"tally_inputs"`
        FilterMode        string                `json:"filter_mode"`  <================== Something like this
	Version           string                `json:"version"`
}
```

PR #290, opened 2024-06-25
feat: filter type Mode

DeshErBojhaa:
## Motivation

Introduce filter type Mode. Also closes this issue https://github.com/sedaprotocol/seda-chain/issues/286

## Testing

Unit tests included.

## Related PRs and Issues

https://github.com/sedaprotocol/seda-chain/issues/286

### DIFF
--- a/x/wasm-storage/keeper/filter.go
+++ b/x/wasm-storage/keeper/filter.go
@@ -1,7 +1,9 @@
 package keeper
 
 import (
+	"encoding/base64"
 	"errors"
+	"fmt"
 
 	"github.com/sedaprotocol/seda-chain/x/wasm-storage/types"
 	"github.com/tidwall/gjson"
@@ -39,8 +41,7 @@ func ApplyFilter(filter []byte, reveals []RevealBody) ([]int, bool, error) {
 			return nil, false, errors.New("empty JSON path in filter input [Mode]")
 		}
 
-		outliers, consensus := filterMode(dataPath, reveals)
-		return outliers, consensus, nil
+		return filterMode(dataPath, reveals)
 
 	// TODO: Reactivate standard deviation filter
 	// case filterStdDev:
@@ -66,13 +67,17 @@ func ApplyFilter(filter []byte, reveals []RevealBody) ([]int, bool, error) {
 // reached in two ways.
 // 1. More than 1/3 of the reveals are corrupted. i.e. has non-zero exit code.
 // 2. More than 2/3 of the data is the same.
-func filterMode(dataPath string, reveals []RevealBody) ([]int, bool) {
+func filterMode(dataPath string, reveals []RevealBody) ([]int, bool, error) {
 	var maxFreq, nonZeroExitCode int
 	freq := make(map[string]int, len(reveals))
 	revealVals := make([]string, 0, len(reveals))
 
 	for _, r := range reveals {
-		val := gjson.Get(r.Reveal, dataPath).String()
+		revealBytes, err := base64.StdEncoding.DecodeString(r.Reveal)
+		if err != nil {
+			return nil, false, fmt.Errorf("filterMode: could not base64 decode: <%s>. Err: %w", r.Reveal, err)
+		}
+		val := gjson.GetBytes(revealBytes, dataPath).String()
 		revealVals = append(revealVals, val)
 		if r.ExitCode != 0 {
 			nonZeroExitCode++
@@ -90,12 +95,12 @@ func filterMode(dataPath string, reveals []RevealBody) ([]int, bool) {
 	// If more than 1/3 reveals have non exit code, we reached a consensus that
 	// these are unusable reveals.
 	if nonZeroExitCode*3 > len(reveals) {
-		return outliers, true
+		return outliers, true, nil
 	}
 
 	// If less than 2/3 matches the max frequency, there is no data-consensus.
 	if maxFreq*3 < len(reveals)*2 {
-		return outliers, false
+		return outliers, false, nil
 	}
 
 	for i, r := range revealVals {
@@ -104,5 +109,5 @@ func filterMode(dataPath string, reveals []RevealBody) ([]int, bool) {
 		}
 		outliers[i] = 0
 	}
-	return outliers, true
+	return outliers, true, nil
 }
--- a/x/wasm-storage/keeper/filter.go
+++ b/x/wasm-storage/keeper/filter.go
@@ -76,6 +76,7 @@ func filterMode(dataPath string, reveals []RevealBody) ([]int, bool) {
 		revealVals = append(revealVals, val)
 		if r.ExitCode != 0 {
 			nonZeroExitCode++
+			continue
 		}
 		freq[val]++
 		maxFreq = max(freq[val], maxFreq)
@@ -92,8 +93,8 @@ func filterMode(dataPath string, reveals []RevealBody) ([]int, bool) {
 		return outliers, true
 	}
 
-	// If not MORE THAN 2/3 matches the max frequency, there is no data-consensus.
-	if maxFreq*3 <= len(reveals)*2 {
+	// If less than 2/3 matches the max frequency, there is no data-consensus.
+	if maxFreq*3 < len(reveals)*2 {
 		return outliers, false
 	}
 
--- a/x/wasm-storage/keeper/filter.go
+++ b/x/wasm-storage/keeper/filter.go
@@ -8,9 +8,8 @@ import (
 )
 
 const (
-	filterNone   byte = 0x00
-	filterMode   byte = 0x01
-	filterStdDev byte = 0x02
+	filterAlgoNone byte = 0x00
+	filterAlgoMode byte = 0x01
 )
 
 // ApplyFilter processes filter of the type specified in the first byte of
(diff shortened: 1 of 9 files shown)


x/wasm-storage/keeper/filter.go
```diff
@@ -20,12 +26,27 @@ func ApplyFilter(filter []byte, reveals []RevealBody) ([]int, bool, error) {
...
+		if dataPath == "" {
+			return nil, false, errors.New("empty JSON path in filter input [Mode]")
+		}
+
+		exitCodes := make([]uint8, 0, len(reveals))
+		revealData := make([]string, 0, len(reveals))
+		for _, r := range reveals {
+			exitCodes = append(exitCodes, r.ExitCode)
+			revealData = append(revealData, r.Reveal)
+		}
```
hacheigriega, line 47: This seems unnecessary to me. I think we should just feed `reveals` directly to `filterMode()`.

x/wasm-storage/keeper/filter.go
```diff
@@ -39,3 +60,55 @@ func ApplyFilter(filter []byte, reveals []RevealBody) ([]int, bool, error) {
...
+	}
+	outliers, consensus := calcOutliers(vals, exitCodes)
+
+	// If more than 1/3 reveals have non exit code, we reached a consensus that
+	// these are unusable reveals.
+	if consensus || nonZeroExitCode*3 > len(reveals) {
+		return outliers, true
+	}
+	return outliers, false
```
hacheigriega, line 81:
```
if nonZeroExitCode*3 > len(reveals) {
    return outliers, true
}

return calcOutliers(vals, exitCodes)
```

x/wasm-storage/keeper/filter.go
```diff
@@ -39,3 +60,55 @@ func ApplyFilter(filter []byte, reveals []RevealBody) ([]int, bool, error) {
 		return outliers, false, nil
 	}
 }
+
+func filterMode(dataPath string, exitCodes []uint8, reveals []string) ([]int, bool) {
```
hacheigriega, line 64: I think we should add a comment about what the function does and what it returns.

hacheigriega (changes requested):

x/wasm-storage/types/codec.go
```diff
@@ -21,3 +25,32 @@ var (
 	Amino     = codec.NewLegacyAmino()
 	ModuleCdc = codec.NewProtoCodec(cdctypes.NewInterfaceRegistry())
 )
+
+const (
+	b             byte   = 0
+	u             uint64 = 0
+	ByteLenFilter        = int(unsafe.Sizeof(b))
+	ByteLenUint64        = int(unsafe.Sizeof(u))
```
hacheigriega, line 33: Not sure about purpose of these when `ByteLenFilter` and `ByteLenUint64` are fixed values.

x/wasm-storage/keeper/filter.go
```diff
@@ -39,3 +60,55 @@ func ApplyFilter(filter []byte, reveals []RevealBody) ([]int, bool, error) {
...
+	for i := 0; i < len(outliers); i++ {
+		outliers[i] = 1
+	}
+
+	var maxFreq int
+
+	for i, r := range reveals {
+		if exitCode[i] != 0 {
+			continue
+		}
```
hacheigriega, line 96: shouldn't the exit code of a reveal treated the same way as the reveal (`r` here)?

x/wasm-storage/types/codec.go
DeshErBojhaa: Initially put 
```
ByteLenFilter        = 1
ByteLenUint64        = 8
```

But it looked ugly. Current approach express the intention more clearly even to a non go reaader.

x/wasm-storage/keeper/filter.go
DeshErBojhaa: not sure what you mean, but I have some code refactored. reopen this if you still want changes.

x/wasm-storage/keeper/filter.go
DeshErBojhaa: here [8c49c59](https://github.com/sedaprotocol/seda-chain/pull/290/commits/8c49c59ff72231ea958fb1dd5e456622d9f10163)

x/wasm-storage/keeper/filter.go
DeshErBojhaa: here [8c49c59](https://github.com/sedaprotocol/seda-chain/pull/290/commits/8c49c59ff72231ea958fb1dd5e456622d9f10163)

x/wasm-storage/keeper/filter.go
```diff
@@ -39,3 +54,54 @@ func ApplyFilter(filter []byte, reveals []RevealBody) ([]int, bool, error) {
...
+	// If more than 1/3 reveals have non exit code, we reached a consensus that
+	// these are unusable reveals.
+	if nonZeroExitCode*3 > len(reveals) {
+		return outliers, true
+	}
+
+	// If not MORE THAN 2/3 matches the max frequency, there is no data-consensus.
+	if maxFreq*3 <= len(reveals)*2 {
+		return outliers, false
+	}
```
mariocao, line 98: I think this should be strictly `<`, so that if we have replication factor 3, then with 1 outlier, there will be consensus. Wdyt?
```suggestion
	// If not MORE THAN 2/3 matches the max frequency, there is no data-consensus.
	if maxFreq*3 < len(reveals)*2 {
		return outliers, false
	}
```

DeshErBojhaa: Now consensus reaches if **at lease** 2/3 is matched.
Before it was **more than** 2/3

[9bae247](https://github.com/sedaprotocol/seda-chain/pull/290/commits/9bae247a8adb32061b10bf4ddc58775c1260ee6c)

x/wasm-storage/keeper/filter.go
```diff
@@ -39,3 +54,55 @@ func ApplyFilter(filter []byte, reveals []RevealBody) ([]int, bool, error) {
...
+// filtering data from the RevealBody->reveal.
+// The function returns a list of int where value of each element can be
+// either 0 or 1. At index i Value 1 means RevealBody[i] contains an
+// outlier. If a reveal have MORE THAN 2/3 frequency, than it's not an
+// outlier.
+//
+// The function also returns if a consensus is reached. Consensus can be
+// reached in two ways.
+// 1. More than 1/3 of the reveals are corrupted. i.e. has non-zero exit code.
+// 2. More than 2/3 of the data is the same.
```
hacheigriega, line 68: My understanding is that consensus is reached as long as more than 2/3 of results, which include both the reveal data and the exit codes, are the same. 
Perhaps we just don't want to distinguish between different non-zero exit codes? 
In any case, I'm confused why we separate the logic for checking for consensus into two: checking the reveal data and checking the exit codes. We can loop through the reveals, counting the frequencies of an occurrence of a given reveal data - exit code pair, and then determine the consensus outcome at once.

x/wasm-storage/keeper/filter.go
```diff
@@ -39,3 +54,55 @@ func ApplyFilter(filter []byte, reveals []RevealBody) ([]int, bool, error) {
...
+func filterMode(dataPath string, reveals []RevealBody) ([]int, bool) {
+	var maxFreq, nonZeroExitCode int
+	freq := make(map[string]int, len(reveals))
+	revealVals := make([]string, 0, len(reveals))
+
+	for _, r := range reveals {
+		val := gjson.Get(r.Reveal, dataPath).String()
```
hacheigriega, line 75: Is there reason why we use `gjson` package here instead of unmarshalling JSON the standard way?

DeshErBojhaa: Yes because the json is unstructured and can get deep. Check the unit tests for some example.

x/wasm-storage/keeper/filter.go
hacheigriega: I see what you mean now. Please ignore the last message.

DeshErBojhaa: For the input 
A0,B1,B1

If you think the output should be
[1,0,0] true;

Then we have different opinion about consensus.

For this input, result should be
[1,1,1] true

All are outliers as we cannot trust 1/3 of the reveals.
And we are in consensus that the output is garbage.




If you are still confused, let's huddle.

x/wasm-storage/types/codec.go
hacheigriega: No big deal, but still unsure what this accomplishes.

hacheigriega (commented):
Nice extensive tests!!

hacheigriega:
> Nice extensive tests!! I think it's good to be merged, but let's get one more approval?

Sorry to retract, but I think we should add test cases to the integration testing `TestTallyVM` so it tests out the whole flow.

DeshErBojhaa:
> Nice extensive tests!! ~I think it's good to be merged, but let's get one more approval?~ EDIT: Sorry to retract, but I think we should add test cases to the integration testing TestTallyVM so it tests out the whole flow.

I've added integration test for MODE filter here: [46c50ba](https://github.com/sedaprotocol/seda-chain/pull/290/commits/46c50ba26511162a7d12104121fd2cd9736975a5)

Let me know if this is sufficient.

hacheigriega (approved):

Issue #291, opened 2024-06-26
♻️ Tally Endblock refactoring


hacheigriega:
## ♻️ Refactor

Address the issues identified during tally implementation:

- [x] First, we have noticed that an error, not just a panic, in the end blocker logic halts the chain with a consensus failure. We need to research into how errors and panics are supposed to be handled in the end blocker and possibly implement a helper to reconcile an end blocker error without halting the chain. Relevant resources:

- https://docs.cosmos.network/main/build/building-apps/security-part-1#in-protocol-panics
- https://github.com/osmosis-labs/osmosis/blob/b0aee0006ce55d0851773084bd7880db7e32ad70/osmoutils/cache_ctx.go#L13-L44

- [x] Emit events during tally process.
- [ ] Check code coverage and add unit tests to achieve a good test coverage. 
  - Make sure to test edge cases.
  - Utilize go mock to intercept and check arguments. See https://pkg.go.dev/github.com/golang/mock/gomock?utm_source=godoc#Call.Do 
  - Add CodeQL analysis to detect and check all possible points of panic in the end blocker.
- [ ] PostDataResults's message structure does not seem to support submission of multiple results at once. (contract needs update - perhaps not a priority)
- [ ] Use limits when fetching tally list from contract.

Thomasvdam:
@hacheigriega does it make sense to move the open points to a new issue and close this one?

hacheigriega:
Agreed. Closing this in favor of #313, which elaborates on the last three tasks.

PR #293, opened 2024-07-03
feat: Handle panic and error in wasm-storage `EndBlock` without halting chain

hacheigriega:
Errors and panics in the wasm-storage module's `EndBlock()` are handled so that they do not cause a chain halt. 

Filtering and tallying logic has been separated into a new function `filterAndTally()` so that an error from this part of the code is recorded as a part of the data result returned to the contract. In such a scenario, the module returns the exit code 255 to signify that the tally VM was not run and that the result message is an error from the module logic.

The `defer` function in `EndBlock()` records an error or a panic from any other part of the end block logic and prevents propagation of the error. 

TODOs
- [ ] Add CodeQL analysis to ensure there is no unexpected points of panics in the end blocker code.
- [x] Add tests to make sure panics and errors are handled correctly (particularly a panic from `filterAndTally()` if this is a possibility).
- [x] Pass unit tests

Related: #291

### DIFF
--- a/x/wasm-storage/keeper/abci.go
+++ b/x/wasm-storage/keeper/abci.go
@@ -93,8 +93,8 @@ func (k Keeper) ProcessTallies(ctx sdk.Context) error {
 	// Loop through the list to apply filter, execute tally, and post
 	// execution result.
 	for _, req := range tallyList {
-		// Construct sudo message to be posted to the contract and
-		// populate the results fields after filterAndTally.
+		// Construct barebone sudo message to be posted to the contract
+		// here and populate its results fields after filterAndTally.
 		sudoMsg := Sudo{
 			ID: req.ID,
 			Result: DataResult{
@@ -109,8 +109,9 @@ func (k Keeper) ProcessTallies(ctx sdk.Context) error {
 
 		vmRes, consensus, err := k.filterAndTally(ctx, req)
 		if err != nil {
-			// Return module error message with exit code 255 to signify
-			// that the tally VM was not executed.
+			// Return with exit code 255 to signify that the tally VM
+			// was not executed due to the error specified in the result
+			// field.
 			sudoMsg.ExitCode = 0xff
 			sudoMsg.Result.ExitCode = 0xff
 			sudoMsg.Result.Result = []byte(err.Error())
--- a/x/wasm-storage/keeper/abci.go
+++ b/x/wasm-storage/keeper/abci.go
@@ -62,7 +62,7 @@ func (k Keeper) ProcessTallies(ctx sdk.Context) error {
 	// Get proxy contract address.
 	contractAddrBech32, err := k.ProxyContractRegistry.Get(ctx)
 	if contractAddrBech32 == "" || errors.Is(err, collections.ErrNotFound) {
-		return fmt.Errorf("proxy contract not reigstered")
+		return fmt.Errorf("proxy contract not registered")
 	}
 	if err != nil {
 		return err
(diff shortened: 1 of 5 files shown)


x/wasm-storage/keeper/abci.go
```diff
@@ -13,17 +13,30 @@ import (
...
+	defer func() {
+		// Handle a panic.
+		if r := recover(); r != nil {
+			k.Logger(ctx).Error("recovered from panic in wasm-storage EndBlock", "err", r)
+		}
+		// Handle an error.
+		if err != nil {
+			k.Logger(ctx).Error("error in wasm-storage EndBlock", "err", err)
+		}
+		err = nil
```
DeshErBojhaa, line 28: Why not propagate the error up to the chain? What happens if we don't set err to nil?

x/wasm-storage/keeper/abci.go
```diff
@@ -13,17 +13,30 @@ import (
 	"github.com/sedaprotocol/seda-wasm-vm/tallyvm"
 )
 
-func (k Keeper) EndBlock(ctx sdk.Context) error {
-	err := k.ProcessExpiredWasms(ctx)
+func (k Keeper) EndBlock(ctx sdk.Context) (err error) {
+	// Use defer to prevent returning an error, which would cause
+	// the chain to halt.
+	defer func() {
+		// Handle a panic.
```
DeshErBojhaa, line 20: Redundant comment.

x/wasm-storage/keeper/abci.go
```diff
@@ -70,83 +84,45 @@ func (k Keeper) ExecuteTally(ctx sdk.Context) error {
...
 			},
-			ExitCode: byte(vmRes.ExitInfo.ExitCode),
 		}
+
+		vmRes, consensus, err := k.filterAndTally(ctx, req)
+		if err != nil {
+			// Return module error message with exit code 255 to signify
+			// that the tally VM was not executed.
+			sudoMsg.ExitCode = 0xff
+			sudoMsg.Result.ExitCode = 0xff
```
DeshErBojhaa, line 115: Is there any case where 
`sudoMsg.ExitCode` and `sudoMsg.Result.ExitCode` will be different?
**If there is no such case**, why duplicate fields?

DeshErBojhaa (approved):
LGTM! With some questions, but not a blocker.

x/wasm-storage/keeper/abci.go
```diff
@@ -177,6 +154,62 @@ func (k Keeper) ExecuteTally(ctx sdk.Context) error {
...
+		return vmRes, consensus, fmt.Errorf("error while applying filter: %w", err)
+	}
+
+	tallyID, err := hex.DecodeString(req.TallyBinaryID)
+	if err != nil {
+		return vmRes, consensus, fmt.Errorf("failed to decode tally ID to hex: %w", err)
+	}
+	tallyWasm, err := k.DataRequestWasm.Get(ctx, tallyID)
+	if err != nil {
+		return vmRes, consensus, fmt.Errorf("failed to get tally wasm for DR ID %s: %w", req.ID, err)
```
Thomasvdam, line 187: Nitpicking but we should either add the DR ID to all errors or leave them out. Given that it's attached to the result for the DR ID I think we can leave it out :)

Thomasvdam (approved):

x/wasm-storage/keeper/abci.go
hacheigriega: The chain halts :)

x/wasm-storage/keeper/abci.go
DeshErBojhaa: Just to understand better, so why we are keeping this `sudoMsg.Result.ExitCode` field?
Isn't it redundant?

Thomasvdam: Right now they're the same, but we want to take a little more time in the future to discuss with Mario how to best handle this field. There's something to be said for keeping `ExitCode` a module level field where for example `255` signals filter failure, and `123` that the VM returned an error, and `Result.ExitCode` is exclusively for the detailed VM error.

PR #299, opened 2024-07-08
feat: Add public key registry module `x/pkr`

### LINKED ISSUE #162: ✨ Pubkeys module
## Motivation

The pubkeys module would serve as a repository for various public keys used in the protocol. Having a separate module for this purpose has the following advantages:

- Provide a central repository of public keys that other modules can have access to
- Facilitate management of keys on various curves (For example, we need to implement a single transaction endpoint for registering keys of any type)
- Support key rotation (to be implemented later)

## Implementation

The store for this module would have the following key-value scheme:
```
validator | curve -> public_key
```

There would be a single transaction endpoint for registering a key. A flag `--rotate` may be added to request overwriting an existing key.
```
$ seda-chaind tx pubkeys register <valoper> <curve> <pubkey> --from <delegator_key>
```

DeshErBojhaa:
# Feature

This PR adds a public key registry module called `x/pkr` that keeps the store of the following key-value schema:
```
validator_address ([]byte) | index (uint32) -> public_key
```

## Explanation of Changes

To generate and register a key, simply specify the index and sign the transaction using the validator operator account. Note the flag `--non-deterministic`, which indicates that the key will be generated from a random entropy.  
```
$ sedad tx pkr add-key 0 --non-deterministic --from satoshi --keyring-backend test --gas auto --gas-prices 10000000000aseda --gas-adjustment 1.8
```
The CLI supports a set of indices and generates a key corresponding to the given index. Initially, only the index `0`, which corresponds to `secp256k1`-curve generated key for VRF, will be supported. The CLI generates a key file in the same directory that contains the private validator file, which by default is `$CHAIN_HOME/config/vrf_key.json`. If the file under this name already exists in this path, the CLI will give an error. Upon successful generation of the key, the CLI will submit a transaction to upload its public key. This transaction must be signed by the validator operator address.

Alternatively, to specify an existing key file, run the transaction with the `--key-file` flag. 
```
$ sedad tx pkr add-key 0 --key-file ~/path/to/vrf_key_file --from satoshi --keyring-backend test --gas auto --gas-prices 10000000000aseda --gas-adjustment 1.8
```

There is also an option to provide a mnemonic phrase. If you specify the `--mnemonic` flag, the CLI will prompt you to input the mnemonic phrase.
```
$ sedad tx pkr add-key 0 --mnemonic --from satoshi --keyring-backend test --gas auto --gas-prices 10000000000aseda --gas-adjustment 1.8
> Enter your bip39 mnemonic
...
```

There is a query endpoint that returns a list of hex-encoded public keys registered for a given validator address:
```
$ sedad query pkr validator-keys sedavaloper1lpkwvawj9y85s6n2dyr7tlt8py9gy3xsleh3cl          

validator_pub_keys:
  pub_keys:
  - index: 0
    pub_key:
      '@type': /cosmos.crypto.secp256k1.PubKey
      key: AmBYlNuBU8up4Y3uE29J+9Lze+U/jQjG/5y2LpXrxHHY
  validator_address: sedavaloper1lpkwvawj9y85s6n2dyr7tlt8py9gy3xsleh3cl
```

## Testing
A couple unit tests were added. More to come in the future.

## Related Issues
Closes #162
Closes https://github.com/sedaprotocol/seda-internal/issues/124


x/pkr/types/gensis.go
mariocao, line 1: Typo in the filename

proto/sedachain/pkr/v1/query.proto
```diff
@@ -0,0 +1,26 @@
...
+import "google/api/annotations.proto";
+
+option go_package = "github.com/sedaprotocol/seda-chain/x/pkr/types";
+
+// Query defines the gRPC querier service.
+service Query {
+  // KeysByApplication returns application specific vrf publick keys.
+  rpc KeysByApplication(KeysByApplicationRequest)
+      returns (KeysByApplicationResponse) {
+    option (google.api.http).get = "/seda-chain/wasm-storage/pkr/keys_by_application";
```
hacheigriega, line 14: remove `wasm-storage`

hacheigriega: Is this http endpoint supposed to take a parameter?

x/pkr/types/tx.go
```diff
@@ -0,0 +1,44 @@
...
+	if !slices.Contains(modules, m.Application) {
+		return ErrInvalidInput.Wrapf("module %s is not present in [%v]", m.Application, modules)
+	}
+	return nil
+}
+
+func (m *MsgAddVRFKey) PublicKey() (cryptotypes.PubKey, error) {
+	pubKey, ok := m.Pubkey.GetCachedValue().(cryptotypes.PubKey)
+	if !ok {
+		return nil, errorsmod.Wrapf(ErrInvalidType, "msg.PublicKey: %T is not a cryptotypes.PubKey", pubKey)
```
hacheigriega, line 35: Is there reason why we use `errorsmod` here only?

x/pkr/keeper/msg_serve.go
hacheigriega, line 1: file name typo

hacheigriega:
I will address the HD-path-based derivation in a separate PR.

x/pkr/keeper/msg_server_test.go
```diff
@@ -0,0 +1,79 @@
...
+			msg: &types.MsgAddKey{
+				ValidatorAddr: valAddrs[0].String(),
+				Index:         0,
+				PubKey:        pubKeys[0],
+			},
+			valAddrBytes: valAddrs[0].Bytes(),
+			wantErr:      nil,
+		},
+		{
+			name: "Fake plastic Any",
```
Thomasvdam, line 38: Plastic?

Thomasvdam (approved):
Probably good to change the target branch to an intermediate branch where we can submit the updates (key generation, interface updates, etc) before merging to main. :)

Issue #284, opened 2024-06-18
✨ Add filter types `Mode` and `StdDev`


hacheigriega:
## ✨ Feature

The chain currently only supports `None` filter, but we hope to add two new filters: `Mode` and `StdDev`. Some preliminary work was done in PR #280 but was removed to prioritize solidifying the minimum tally requirement first. Both the corresponding unit testing and integration testing should be updated to cover these new types.

DeshErBojhaa:
Mode is implemented here https://github.com/sedaprotocol/seda-chain/pull/290

PR #300, opened 2024-07-09
feat: Refactor filtering and adding std dev filter

hacheigriega:
## Explanation of Changes

Fixing issues identified during devnet testing

- Consensus is always set to false if tally VM does not run even if filtering result gives consensus = true.
- Handle the case when reveal value is not found at a given JSON path.
- Invalid JSON path, invalid reveal encoding, and non-zero exit code should all be treated as "corrupt reveals."
- Add events
- Hex-encode tally VM inputs.

This PR also refactors filtering and adds a new filter standard deviation.

- Make use of a new interface type `Filter` to clean up the code.
- Add standard deviation filter

Closes: #284 
Closes: #286


x/wasm-storage/keeper/filter.go
```diff
@@ -56,58 +55,65 @@ func ApplyFilter(filter []byte, reveals []RevealBody) ([]int, bool, error) {
...
+	// If less than 2/3 of the reveals match the max frequency,
+	// there is no consensus.
 	if maxFreq*3 < len(reveals)*2 {
 		return outliers, false, nil
 	}
 
 	for i, r := range revealVals {
-		if freq[r] != maxFreq || reveals[i].ExitCode != 0 {
-			continue
+		if freq[r] == maxFreq {
```
hacheigriega, line 114: My reasoning was that since we know at this point that

- Corrupt reveals are <= 1/3
- Max frequency occurs >= 2/3

=> corrupt reveals can never have the frequency of `maxFreq`.

Thomasvdam (commented):
At a high level the code makes sense to me (as names of functions, location in files, etc). I need to take a real close look at the filters: how they determine consensus and if this aligns with what Franklin and Mario have in mind. I'm not sharp enough right now, so I'll have a look tomorrow together with Franklin.

hacheigriega:
@Thomasvdam  Sounds good and thanks. I'll have this PR fully ready for review by then. I will also share a flow chart I drew.

hacheigriega:
The specs are outlined here https://www.notion.so/sedaprotocol/Tallying-9388fd590d1c4b659abb721b57909feb?pvs=4#114307cdfc9d4c8b929dc11deb540ca3

x/wasm-storage/types/filters.go
```diff
@@ -0,0 +1,299 @@
...
+			j++
+			continue
+		}
+		if isOutlier(maxSigma, nums[i], median, medianHalf) {
+			outliers[i+j] = 1
+		} else {
+			nonOutlierCount++
+		}
+		i++
+	}
```
FranklinWaller, line 254: I'm having a lil bit of a hard time understanding what is going on here. Can we maybe simplify this part (It doesn't matter if its less efficient)

x/wasm-storage/types/filters.go
```diff
@@ -0,0 +1,299 @@
...
+		return false
+	}
+
+	if diff > (maxSigma / 1e6) {
+		return true
+	} else if medianHalf && diff == (maxSigma/1e6) {
+		// Means that diff = int(maxSigma) + 0.5
+		// so now check that maxSigma's decimal part is > 0.5
+		// by checking that last 6 digits of maxSigma
+		return maxSigma%1e6 > 5e6
```
FranklinWaller, line 284: I think this should be 5e5 not 5e6?

x/wasm-storage/types/filters.go
```diff
@@ -0,0 +1,299 @@
...
+
+	// If less than 2/3 of the numbers fall within max sigma range
+	// from the median, there is no consensus.
+	if nonOutlierCount*3 < len(nums)*2 {
+		return outliers, ErrNoConsensus
+	}
+	return outliers, nil
+}
+
+func isOutlier[T constraints.Integer](maxSigma uint64, num, median T, medianHalf bool) bool {
```
FranklinWaller, line 264: I think we should have one function that calculates the Sigma (in a seperate file) that doesn't know about reveals or protocol logic (apart from the 6 decimals). And together with the medianHalf logic. Right now its a lil splinters across functions while the logic is one thing as a whole.

FranklinWaller (changes requested):

x/wasm-storage/types/filters.go
```diff
@@ -0,0 +1,299 @@
...
+		}
+		if len(bz) != int(rt.Size()) {
+			corruptCount++
+			corruptInds = append(corruptInds, i)
+			continue
+		}
+
+		switch rt.Kind() {
+		case reflect.Int32:
+			var num int32
```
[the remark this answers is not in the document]
hacheigriega: Really good point. I took this approach because originally I thought we had to use different functions depending on the type, but it seems we can generalize the code and use T.

x/wasm-storage/types/filters.go
hacheigriega: Thank you !!!!!!

hacheigriega: will also add some relevant tests

hacheigriega: also have to flip `>`

x/wasm-storage/types/filters.go
```diff
@@ -0,0 +1,299 @@
...
+	rt := reflect.TypeOf(z)
+	length := len(dataList)
+	for i, data := range dataList {
+		if data == "" {
+			corruptCount++
+			corruptInds = append(corruptInds, i)
+			continue
+		}
+
+		bz, err := base64.StdEncoding.DecodeString(data)
```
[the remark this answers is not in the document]
hacheigriega: `parseReveal` decodes the reveal and retrieves the JSON value at the given data path, but this value is still base64-encoded.

x/wasm-storage/types/filters.go
hacheigriega: I feel bad for making you review this now 😬😬 I added `Sigma` type in [bb15b42](https://github.com/sedaprotocol/seda-chain/pull/300/commits/bb15b42eedf6d60aebc92a204ce84a7db336450b) and `HalfStepInt` in [22a571d](https://github.com/sedaprotocol/seda-chain/pull/300/commits/22a571d2a4c7ed40e0ec2fc04e7898f62c6c0c40)

x/wasm-storage/types/sigma.go
```diff
@@ -0,0 +1,94 @@
...
+		if h.integer < 0 && h.halfStep && absDiff > 0 {
+			absDiff--
+		}
+	case h.integer < x:
+		absDiff = uint64(x - h.integer)
+		// Note absDiff > 0 check is unnecessary.
+		if h.integer >= 0 && h.halfStep {
+			absDiff--
+		}
+	}
```
hacheigriega, line 82: Let me reexamine the logic here. I realized the comments may not be correct.

hacheigriega: updated comments

hacheigriega: fixed comments and added `neg` field to `HalfStepInt` to represent -0.5.

hacheigriega:
Ready for final review. The simulation test fails due to its inherent instability. We will have to wait for Cosmos team's simulation test refactoring.

x/wasm-storage/keeper/filter_test.go
```diff
@@ -49,22 +51,50 @@ func TestFilterNone(t *testing.T) {
...
+			reveals: []types.RevealBody{
+				{Reveal: `{"result": {"text": "A", "number": 0}}`},
+				{Reveal: `{"resultt": {"text": "A", "number": 10}}`},
+				{Reveal: `{"result": {"text": "A", "number": 101}}`},
+			},
+			consensus: true,
+			wantErr:   nil,
+		},
 		{
 			name:            "Mode filter - Consensus due to non exit code",
```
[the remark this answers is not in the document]
hacheigriega: the second reveal is corrupt :D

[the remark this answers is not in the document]
Thomasvdam: Apologies, I meant that I think it should say there is no consensus because it now implies the expectation is for there to be consenus 😬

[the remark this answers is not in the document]
hacheigriega: THere is consensus though bc 2/3 is non-outlier?

x/wasm-storage/types/sigma.go
```diff
@@ -0,0 +1,94 @@
...
+func (h *HalfStepInt[T]) Mid(x, y T) *HalfStepInt[T] {
+	h.integer = (x + y) / 2
+	h.neg = (x + y) < 0
+	h.halfStep = false
+	// Set h's halfStep to true if the earlier division operation has
+	// truncted the result.
+	if (x%2 == 0 && y%2 != 0) ||
+		(x%2 != 0 && y%2 == 0) {
+		h.halfStep = true
+	}
```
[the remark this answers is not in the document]
hacheigriega: I ended up just adding a comment because I didn't feel it was necessary to add a function.

hacheigriega:
@Thomasvdam Oh yes you're right. I thought we were assuming that the results would be base64-encoded, but I realized that I came up with that assumption myself. We should probably use JSON number for this instead, huh?

Thomasvdam:
Yeah I think so :)

hacheigriega:
> Yeah I think so :)

Fixed in [482764d](https://github.com/sedaprotocol/seda-chain/pull/300/commits/482764d6007d70fffc12c61ad66992fcedf600e4). `parseReveals()` parses any values into strings, which then are parsed into integers using `strconv` functions in the standard deviation filter. So for example both `4` and `"4"` work.

x/wasm-storage/types/errors.go
```diff
@@ -3,7 +3,14 @@ package types
...
-	ErrAlreadyExists = errors.Register("wasm-storage", 2, "already exists")
-	ErrInvalidLen    = errors.Register("wasm-storage", 3, "invalid len")
+	ErrInvalidParam          = errors.Register("wasm-storage", 1, "invalid param")
+	ErrAlreadyExists         = errors.Register("wasm-storage", 2, "already exists")
+	ErrInvalidFilterType     = errors.Register("wasm-storage", 3, "invalid filter type")
+	ErrInvalidFilterInputLen = errors.Register("wasm-storage", 4, "invalid filter input length")
+	ErrInvalidPathLen        = errors.Register("wasm-storage", 5, "invalid JSON path length")
+	ErrEmptyReveals          = errors.Register("wasm-storage", 6, "no reveals given")
+	ErrCorruptReveals        = errors.Register("wasm-storage", 7, "more than 1/3 of the reveals are corrupted")
+	ErrNoConsensus           = errors.Register("wasm-storage", 8, "less than 2/3 of the reveals in consensus range")
```
mariocao, line 13:
```suggestion
	ErrNoConsensus           = errors.Register("wasm-storage", 8, "More than 1/3 of the reveals are not in consensus range")
```

x/wasm-storage/types/filters.go
```diff
@@ -0,0 +1,235 @@
...
+// NewFilterMode constructs a new FilerMode object given a filter
+// input.
+// Mode filter input looks as follows:
+// 0             1                  9       9+data_path_length
+// | filter_type | data_path_length |   data_path   |
+func NewFilterMode(input []byte) (FilterMode, error) {
+	var filter FilterMode
+	if len(input) < 9 {
+		return filter, ErrInvalidFilterInputLen.Wrapf("%d < %d", len(input), 9)
+	}
```
FranklinWaller, line 46: Do we have a test for when you input a valid length but an empty string as data_path? 
According to the JSON path spec you need at least `$` in order to count as a valid data_path. But if it's handled correctly it should be fine.

FranklinWaller: After a lil discussion with @mariocao it seems that the case with an empty data_path is erroring in the parsing of the JSON + path. However it would be better if we errored before and not execute the tally at all.

x/wasm-storage/types/filters_util.go
```diff
@@ -0,0 +1,59 @@
...
+	var maxFreq, corruptCount int
+	freq := make(map[string]int, len(reveals))
+	dataList := make([]string, len(reveals))
+	for i, r := range reveals {
+		if r.ExitCode != 0 {
+			corruptCount++
+			continue
+		}
+
+		revealBz, err := base64.StdEncoding.DecodeString(r.Reveal)
```
FranklinWaller, line 35: What is revealBz?

x/wasm-storage/types/sigma.go
```diff
@@ -0,0 +1,94 @@
...
+func NewSigma(data []byte) (Sigma, error) {
+	var s Sigma
+	err := binary.Read(bytes.NewReader(data), binary.BigEndian, &s)
+	if err != nil {
+		return s, err
+	}
+	return s, nil
+}
+
+// WholeNumber returns Sigma's whole number part.
```
FranklinWaller, line 23: Maybe add a comment, something like `1_000_000` becomes `1` (and same with fractional part). Just for clarity

x/wasm-storage/types/filters.go
```diff
@@ -0,0 +1,235 @@
...
+	case 0x00: // Int32
+		return detectOutliersInteger[int32](dataList, f.maxSigma)
+	case 0x01: // Int64
+		return detectOutliersInteger[int64](dataList, f.maxSigma)
+	case 0x02: // Uint32
+		return detectOutliersInteger[uint32](dataList, f.maxSigma)
+	case 0x03: // Uint64
+		return detectOutliersInteger[uint64](dataList, f.maxSigma)
+	default:
+		return nil, ErrInvalidNumberType
```
FranklinWaller, line 152: Maybe we should create an issue for the big int so we can add this later

x/wasm-storage/types/errors.go
hacheigriega: "More than or equal to 1/3 of the reveals are not in consensus range" right?

x/wasm-storage/types/filters.go
hacheigriega: If the data_path is invalid, tally won't be executed due to corrupt reveal error. You're right that we don't validate the data_path except its length until we parse the reveals, but I don't think we can do anything without looking at the reveal data?

x/wasm-storage/types/filters_util.go
hacheigriega: Cosmos uses `bz` for some reason for byte slices, but I realized it's just a Cosmos thing :D Will update this

x/wasm-storage/types/filters.go
hacheigriega: Weirdly enough, our gjson package fails when you put `$` in the beginning of a JSON path.

hacheigriega: The package we use for json path search seems to use a different variant of path syntax https://github.com/tidwall/gjson/blob/master/SYNTAX.md
Will check if there is a good a alternative

x/wasm-storage/types/errors.go
mariocao: True 🤣

x/wasm-storage/types/filters.go
FranklinWaller: Sorry ignore my comment about the $. Its indeed not a strict requirement. The gjson syntax is indeed not the same as JSONPath

Thomasvdam: I think using JSONPath makes sense though, it's the closest we have to a standard for this kind of thing. Looks like there are some good Golang implementations

hacheigriega: Agree with @Thomasvdam but this change is not as trivial as I thought. GJSON was quite convenient for what we want, which is to just get string representations of JSON values at a given path. (We later parse these strings to different data types)

hacheigriega: Chose the package [OjG](https://github.com/ohler55/ojg) for standardized JSONPath as described [here](https://goessner.net/articles/JsonPath/).

x/wasm-storage/keeper/abci.go
```diff
@@ -149,15 +161,26 @@ func (k Keeper) ProcessTallies(ctx sdk.Context) error {
...
 		)
+		ctx.EventManager().EmitEvent(
+			sdk.NewEvent(
+				types.EventTypeTallyCompletion,
+				sdk.NewAttribute(types.AttributeRequestID, req.ID),
+				sdk.NewAttribute(types.AttributeTypeConsensus, strconv.FormatBool(consensus)),
+				sdk.NewAttribute(types.AttributeTallyVMStdOut, strings.Join(vmRes.Stdout, "\n")),
+				sdk.NewAttribute(types.AttributeTallyVMStdErr, strings.Join(vmRes.Stderr, "\n")),
+			),
+		)
```
Thomasvdam, line 172: Small favour, can we add the `exit_code` attribute here which has the exit code of the tally module?

hacheigriega: done!

Thomasvdam (approved):

x/wasm-storage/types/events.go
```diff
@@ -3,7 +3,14 @@ package types
...
 	EventTypeStoreDataRequestWasm = "store_data_request_wasm"
 	EventTypeOverlayWasm          = "store_overlay_wasm"
+	EventTypeTallyCompletion      = "tally_completion"
+	EventTypeWasmExpiration       = "wasm_expiration"
 
-	AttributeHash     = "hash"
-	AttributeWasmType = "wasm_type"
+	AttributeWasmHash      = "wasm_hash"
+	AttributeWasmType      = "wasm_type"
+	AttributeRequestID     = "request_id"
```
Thomasvdam, line 11: Sorry to keep coming back for small things but maybe we should make this `dr_id`? That's what the contract uses and consistency is nice 😬

hacheigriega: no problem. updated!

FranklinWaller (approved):

Issue #303, opened 2024-07-15
🐛 Empty event from wasm_storage module


Thomasvdam:
## 🐛 Bug Report

The custom event emitted during the handling of an upload WASM message does not contain any attributes or even a type. See https://sedaprotocol.github.io/test-explorer/devnet/tx/6A455A405F5972C7F8633B9026609E71F1FBA6605D459A4A73CCB1BDBBFC3AF0
or https://devnet.test.explorer.seda.xyz/txs/6A455A405F5972C7F8633B9026609E71F1FBA6605D459A4A73CCB1BDBBFC3AF0

In addition I think we can probably skip emitting the bytecode in the event given that we already have it in the message input, but I can also see the value of keeping it there in case others want to build indexing software based on just the events...

## Steps to Reproduce

1. Upload a WASM binary to the chain.
2. Check the events of the transaction.

#### Stack trace & error message

We get a weird empty event without a type or attributes.

```json
{
  "attributes": [
    {
      "index": true,
      "key": "msg_index",
      "value": "0"
    }
  ],
  "type": ""
}
```

## Expected Behavior

I'd expect the events to contain the information that was emitted during the execution of the message.

```json
{
  "attributes": [
    {
      "index": true,
      "key": "msg_index",
      "value": "0"
    },
    {
      "index": true,
      "key": "hash",
      "value": "HASH_HERE"
    },
    {
      "index": true,
      "key": "wasm_type",
      "value": "ENUM_HERE"
    }
  ],
  "type": "some-event-type"
}
```

## Your Environment

- sedad: `v0.2.0-dev.2`
- env: devnet

PR #305, opened 2024-07-18
fix: fix empty event bug from wasm-storage module

99adarsh:
## Explanation of Changes
- Removed `EventStoreDataRequestWasmWrapper` and `EventStoreOverlayWasmWrapper` wrappers for emitting events
- Directly using `EventStoreDataRequestWasm` and `EventStoreOverlayWasm` proto event messages fixes the issue

Closes: #303


hacheigriega:
Thanks for the PR!! So these wrappers were introduced in #135 to remove unnecessary double quotes. Ideally we would like to fix this without undoing #135.

hacheigriega:
We should also address `TestMarshalJSON()` in `msg_server_test.go` if we make changes in the event types.

Issue #279, opened 2024-06-03
♻️ Wasm type refactor


hacheigriega:
## ♻️ Refactor

Merge two Wasm types `WASM_TYPE_DATA_REQUEST` and `WASM_TYPE_TALLY`. 
Also consider removing the type `WASM_TYPE_UNSPECIFIED`.

Issue #277, opened 2024-05-31
♻️ Reflect change to mono contract 


hacheigriega:
## ♻️ Refactor

The switch to mono contract calls for a few chain-side changes:
- Rename proxy contract instantiation to SEDA contract instantiation.
- Update contract deployment scripts accordingly.

PR #308, opened 2024-07-23
refactor: Various refactoring chores

hacheigriega:
## Explanation of Changes

- Proxy Contract has been renamed to Core Contract
- Removed remnants of staking and data request contracts as we migrate to mono contract structure
- Wasm types have been refactored. There are now three wasm types: Data request (which includes both data request and tally wasms), data request executor, and relayer. Note that the transaction `tx wasm-storage store-data-request-wasm` no longer takes a wasm type flag.
- 

## Related PRs and Issues

Closes: #279
Closes: #277


proto/sedachain/wasm_storage/v1/wasm_storage.proto
```diff
@@ -22,19 +22,14 @@ message Wasm {
...
-  WASM_TYPE_TALLY = 2 [ (gogoproto.enumvalue_customname) = "WasmTypeTally" ];
-  // A wasm that is an overlay executor.
-  WASM_TYPE_DATA_REQUEST_EXECUTOR = 3
+  // WASM_TYPE_DATA_REQUEST_EXECUTOR is for overlay executors.
+  WASM_TYPE_DATA_REQUEST_EXECUTOR = 1
       [ (gogoproto.enumvalue_customname) = "WasmTypeDataRequestExecutor" ];
-  // A wasm that is an overlay relayer.
-  WASM_TYPE_RELAYER = 4
+  // WASM_TYPE_RELAYER is for overlay relayers.
+  WASM_TYPE_RELAYER = 2
```
Thomasvdam, line 32: @mariocao @FranklinWaller does it still make sense to have this as part of the protocol if we're focussing more on custom relayers?

x/wasm-storage/types/wasm.go
```diff
@@ -51,13 +51,11 @@ func WasmTypeFromString(s string) WasmType {
...
 		return WasmTypeDataRequest
-	case "TALLY":
-		return WasmTypeTally
 	case "DATA-REQUEST-EXECUTOR":
 		return WasmTypeDataRequestExecutor
 	case "RELAYER":
 		return WasmTypeRelayer
 	default:
-		return WasmTypeNil
+		panic("unexpected wasm type")
```
Thomasvdam, line 59: I guess this is safe as we're only using it in the CLI? Is there some way we can enforce that so we don't run into issues if for whatever reason later down the line we need this functionality in the chain and whoever implements it doesn't realise the function panics.

Thomasvdam (commented):
One minor thing, but want to think more about the relayer WASM before approving.

proto/sedachain/wasm_storage/v1/wasm_storage.proto
FranklinWaller: I don't think it does anymore

mariocao: Agree. 👍

x/wasm-storage/types/wasm.go
hacheigriega: Yes, and it's ok to use it in txs because they're isolated. You're right that we should be careful when it comes to non-tx-initiated application logic though.

I don't think there is a way to enforce that at the code level, but we can look into some code analysis engines. It is mentioned in this issue, but I will elaborate more https://github.com/sedaprotocol/seda-chain/issues/291#issue-2376044228

proto/sedachain/wasm_storage/v1/wasm_storage.proto
hacheigriega: done!

Thomasvdam (approved):

PR #315, opened 2024-08-07
feat: Support HD path for key generation in `x/pkr`

hacheigriega:


hacheigriega:
We decided against hierarchical derivations of SEDA keys.

Issue #298, opened 2024-07-05
🐛 Static ARM build crashes on tally


Thomasvdam:
## 🐛 Bug Report

When the node encounters a block where the tally VM is invoked the process stops with a core dump on the first call into the tally VM.

This does not happen with a dynamically linked build of the exact same version, built by checking out the tag and running `make build`.

## Steps to Reproduce

Set up the node as usual and attempt to sync up with a network that has DRs with enough reveals to proceed to the tally stage.

#### Stack trace & error message

Logs taken from journalctl:
```
Jul 04 08:24:41 cosmovisor[36022]: 8:24AM INF received proposal module=consensus proposal="Proposal{3113/0 (92CE960CC2FF34095E6F573900059B3CD29C4252C89FB4E3CF4EFD2014BBE2A8:1:E0A7869C6087, -1) 18C405B66B5D @ 2024-07-04T08:24:41.662000716Z}" proposer=470887968BE1D4E8370874AE0558C6C68AF70A25
Jul 04 08:24:41 cosmovisor[36022]: 8:24AM INF received complete proposal block hash=92CE960CC2FF34095E6F573900059B3CD29C4252C89FB4E3CF4EFD2014BBE2A8 height=3113 module=consensus
Jul 04 08:24:41 cosmovisor[36022]: 8:24AM INF finalizing commit of block hash=92CE960CC2FF34095E6F573900059B3CD29C4252C89FB4E3CF4EFD2014BBE2A8 height=3113 module=consensus num_txs=1 root=A2B6050363AC67CB8A36B511A80EB96B77DE88BC915F01176409F8EA02B71F3B
Jul 04 08:24:41 cosmovisor[36022]: 8:24AM INF non-empty tally list - starting tally process module=x/wasm-storage
Jul 04 08:24:41 cosmovisor[36022]: 8:24AM INF executing tally VM arguments=["","[{\"reveal\":[84,104,105,115,32,105,115,32,97,32,68,82,32,101,120,101,99,117,116,105,111,110,32,114,101,115,117,108,116],\"salt\":[],\"exit_code\":0,\"gas_used\":\"0\"}]","[0]"] consensus=true module=x/wasm-storage request_id=88c7c0bf0f22b71328fccc2180871f600f50733565afe3d9bfb80f46753fbd9d tally_wasm_hash=e728da7f1457fcf1110f8664e1962d0076d1f301cef09e5f130b9a7d60ccef06
[...]

PR #321, opened 2024-08-13
fix: Adding build tag for Rosetta

hacheigriega:
## Motivation

Adding an optional build tag called `rosetta` to determine whether to add Rosetta support or not.
To build with this tag using make, run:
```
ENABLE_ROSETTA=true make build
```

Related: #298 
Thank you @Thomasvdam for the solution!

### DIFF
--- a/cmd/sedad/cmd/rosetta_disabled.go
+++ b/cmd/sedad/cmd/rosetta_disabled.go
@@ -1,0 +1,11 @@
+//go:build !rosetta
+
+package cmd
+
+import (
+	"github.com/spf13/cobra"
+
+	"github.com/sedaprotocol/seda-chain/app"
+)
+
+func addRosettaCmd(_ *cobra.Command, _ app.EncodingConfig) {}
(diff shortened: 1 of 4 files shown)


Thomasvdam (approved):
It feels a little clunky but I think this is the way to Go 👉 😎 👉

Thomasvdam:
Actually we might want to add a note somewhere that rosetta and static can't go together 🤔 Maybe in the Makefile and have it error when both tags are present?

Or we add require the static tag to not be set for the Rosetta file, I think you can do boolean logic with tags.

cmd/sedad/cmd/rosetta_disabled.go
```diff
@@ -0,0 +1,11 @@
+//go:build !rosetta
```
Thomasvdam, line 1: We can make this

```golang
//go:build !rosetta static
```

And the other

```golang
//go:build rosetta,!static
```

Provided there is a static build tag... 😬

hacheigriega:
Added an error message when both `ENABLE_ROSETTA` and `LINK_STATICALLY` are set to true.
```
LINK_STATICALLY=true ENABLE_ROSETTA=true make build                                                      
Makefile:78: *** Cannot link statically when Rosetta is enabled.  Stop.
```

gluax (approved):

Issue #316, opened 2024-08-12
✨ Data Proxy Registry


Thomasvdam:
## Motivation

The data proxy registry module will serve as a repository for public keys and the fee of the associated data proxies.

After executing a `proxy_http_fetch` method an overlay node will query the data proxy registry with the public key it received from the upstream data proxy node. It can use the params to determine the amount of gas that was consumed by the `proxy_http_fetch` call.

> [!WARNING]
> We will want to implement a delay to fee updates to prevent attacks and give consumers a chance to respond to fee changes.

## Implementation

The store for this module would have the following key-value scheme:
```
[map]  public_key -> data_provider_params
# Delayed update queue
[set]  height/timestamp | public_key
[item] module_params
```

Where `data_provider_params` are the settings to be used for that provider:
- Payout address: a seda address that receives the rewards for providing data-proxy services.
- Fee in `aseda`: the cost of the data-proxy service. This can be converted to gas using the gas price of a data request.
- Upcoming fee: a data-proxy operator can schedule a fee change to give consumers the opportunity to edit their data requests. Null when no fee change is scheduled. Otherwise a struct that includes the new fee and the height at which it comes into effect. 

`module_params` should at least include:
- Fee update delay: a positive integer that specifies how many blocks need to pass before a fee change comes into effect.

There should be 2 transaction endpoints. One for registering a key with params, and another to edit the params for an existing public key.
```sh
$ sedad tx dataproxy register <pubkey> <params> --from <address>
# Updates the payout address right away. If a new fee is specified it will be scheduled according to the delay in the module params, unless a manual delay is specified (which should be higher than the minimum).
$ sedad tx dataproxy edit <pubkey> <params> --from <address> [--delay int64]
```

Scheduling a fee update while there is already a pending fee update overwrites the old one with the new update and calculates a new effective height based on the delay.
This should also remove the old entry from the queue.

There should be a query endpoint to retrieve the params for a particular public key.
```sh
$ sedad query dataproxy proxy-params <pubkey>
```

PR #322, opened 2024-08-15
Initial module boilerplate

Thomasvdam:
## Motivation

Splitting this up into several PRs to try and keep it reviewable... This one's already pretty big and contains a little bit of logic, in hindsight I should have started with an empty module rather than taking the PKR work and changing that until the linter and compiler stopped yelling at me.

I'd appreciate feedback on the following areas in particular:
- All the protobuf files
- Naming of the types in the protobuf files
- Name of the module directory (`"data-proxy-registry"`)
- Name of the module (`"data-proxy-registry"`)
- Name of the package (`"dataproxyregistry"`)

## Explanation of Changes

Mainly boilerplate code in terms of Cosmos side implementation, with the exception of the Params message handling, some message validation basics, and part of the collections on the keeper.

Most of the protobuf definitions are close to what they will probably end up being. However I'm sure I'm misusing some of the annotations with regards to nullability, that will most likely change once I start working with the types and figure out what all the annotations do exactly.

## Testing

None, there is barely any implementation that's worth testing by itself, it should all be covered once we start implementing the actual features and testing those.

## Related PRs and Issues

Part-of: #316


x/data-proxy/types/keys.go
```diff
@@ -0,0 +1,13 @@
...
+import "cosmossdk.io/collections"
+
+const (
+	ModuleName = "data-proxy"
+	StoreKey
+)
+
+var (
+	DataProxyConfigPrefix = collections.NewPrefix(0)
+	ParamsPrefix          = collections.NewPrefix(3)
```
hacheigriega, line 12: `collections.NewPrefix(1)`, right?

x/data-proxy/types/tx.go
```diff
@@ -0,0 +1,31 @@
+package types
+
+import (
+	sdk "github.com/cosmos/cosmos-sdk/types"
+)
+
```
hacheigriega, line 6: In case you forget (because I usually forget), we will probably have to add `UnpackInterfaces` for message types that involve `Any` types so we can unwrap proto to the go type.

hacheigriega (approved):

x/data-proxy/types/tx.go
Thomasvdam: I think with the move to string this is not relevant anymore. :)

PR #323, opened 2024-08-15
feat: Post data results in batch and lint imports

### LINKED ISSUE #313: ♻️ Improve module - contract interaction in x/tally
## ♻️ Refactor

There are at least two ways the current implementation of module - Core Contract interaction in the tally module can be improved.

First, when fetching tally list from the Core Contract, the query message does not make use of parameters like limit or offset. Making use of these parameters could improve the robustness and safety of the code.

Second, the tally module currently submits `PostDataResults` back to the contract one by one after each tally VM execution. Since the Core Contract has been updated to take an array of this `Sudo` messages at once, we can make use of this update to only make one `Sudo` call per block.

Overall testing involving module - contract interaction can be improved as well. Since it is difficult to configure the Core Contract state so that it expects a `PostDataResults` `Sudo` call, we are only checking if the contract returns an expected error. Perhaps in addition to this check, we can make use of go mock to intercept and examine the arguments being passed to the `Sudo` call as well (See https://pkg.go.dev/github.com/golang/mock/gomock?utm_source=godoc#Call.Do). Also, check if CodeQL analysis can be used to detect any overlooked possible points of panic in end blocker.

hacheigriega:
## Explanation of Changes

- Post tally execution results to the Core Contract in batch rather than one by one.
- Organize imports by adding import groupings to the linter.

Closes #313


gluax (approved):

x/tally/keeper/testdata/contracts.go
Thomasvdam, line 1: Are these changes part of this PR?

Thomasvdam (approved):

x/tally/keeper/testdata/contracts.go
hacheigriega: Yeah.. forgot to move these files used in tally tests when I created the tally module..

PR #324, opened 2024-08-16
Add registration message handling

Thomasvdam:
## Motivation

Also introduces an admin address to make updates to the proxy config easier for operators. An alternative would be to use a sequence number, but that makes the whole managamenet process more cumbersome for the operator.

## Explanation of Changes

- Wire the module into the app so it's possible to test the CLI
- Add CLI command for the registration transaction (although we expect these to be done through the explorer)
- Add message handling for the registration

## Testing

Added tests for the message server. If there are other tests that make sense at this stage I would love to learn!

> [!WARNING]
> The happy_path_with_memo test fails since we don't have a test vector where a message with memo was signed.

## Related PRs and Issues

Part-of: #316


gluax (commented):
The simulation test doesn't look like it calls the failing unit test, so I think we need to update that somehow as well? Otherwise LGTM.

app/app.go
```diff
@@ -132,6 +132,9 @@ import (
 	"github.com/sedaprotocol/seda-chain/app/keepers"
 	appparams "github.com/sedaprotocol/seda-chain/app/params"
 	"github.com/sedaprotocol/seda-chain/docs"
+	dataproxy "github.com/sedaprotocol/seda-chain/x/data-proxy"
+	dataproxyKeeper "github.com/sedaprotocol/seda-chain/x/data-proxy/keeper"
+	dataproxyTypes "github.com/sedaprotocol/seda-chain/x/data-proxy/types"
```
hacheigriega, line 137: Can't find the doc but I think the convention is to keep package import names in lower case. So `dataproxykeeper` and `dataproxytypes`

hacheigriega (approved):

PR #325, opened 2024-08-16
chore(data-proxy): update proxy query not found error

Thomasvdam:
## Motivation

Give the user better feedback when a public key does not have an associated data proxy config registered.

## Explanation of Changes

Not much, just updating what was already there from the boilerplate, adding tests, and making the error a little more user friendly when an unknown public key is queried.

## Testing

Added tests for the query server.

## Related PRs and Issues

Part-of: #316

### DIFF
--- a/x/data-proxy/keeper/common_test.go
+++ b/x/data-proxy/keeper/common_test.go
@@ -3,7 +3,10 @@ package keeper_test
 import (
 	"testing"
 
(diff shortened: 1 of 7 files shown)


gluax (approved):

gluax:
I approved, but I wasn't sure if the tests were supposed to fail on this one.

hacheigriega (approved):

PR #326, opened 2024-08-16
docs: Add Swagger docs

### LINKED ISSUE #159: 🐛 Broken REST Routes
## Overview

There's at least one REST route that's currently broken. Likely more. Some more urgent, some less. We need to fix them based on priority:
- `/cosmos/tx/v1beta1/txs?events=message.sender='seda...'`

## Investigation
After some investigation the `/txs` route broke exactly in commit [`45395245601a363af836a3bb4a59a9385d4991cb`](https://github.com/sedaprotocol/seda-chain/commit/45395245601a363af836a3bb4a59a9385d4991cb). 

### Before Upgrading to SDK v0.50
`/cosmos/tx/v1beta1/txs?events=message.sender='seda...'` worked normally.
```
http://localhost:1317/cosmos/tx/v1beta1/txs?events=message.sender='seda1e32g7uv8pw7jrxdhtz7adxfws0vw6u9gzcxccq'
```
✅ image

### After Upgrading to SDK v0.50
Check out the commit prior, i.e. [`b4b6ed7acd017ebbccfe1a62cbdec053275a6847`](https://github.com/sedaprotocol/seda-chain/commit/b4b6ed7acd017ebbccfe1a62cbdec053275a6847). And you will see the route no longer works.
```
http://localhost:1317/cosmos/tx/v1beta1/txs?events=message.sender='seda1e32g7uv8pw7jrxdhtz7adxfws0vw6u9gzcxccq'
```
🔴 image

hacheigriega:
## Motivation

Adding Swagger docs for all query points. To enable it, set `swagger = true` in `app.toml` and go to http://localhost:1317/swagger/

To re-generate Swagger docs, run `make proto-swagger-gen`. Note that `client/docs/config.json` has to be updated to add or delete query endpoints from the doc.

Closes #159


Thomasvdam:
This feels pretty intense, but then again this is often the case with code/docs generation. 

Why do we have to manually check out all dependencies and move their proto files? I tried looking at a few other projects (Gaia and Osmosis) but it seems they don't need/do that?

Thomasvdam (approved):
Aaaah I didn't check beyond their root makefile, that explains :)

PR #327, opened 2024-08-17
Handle edits to the data proxy config

Thomasvdam:
## Motivation

Allow data proxy operators to update their data proxy config on chain.

## Explanation of Changes

I purposefully didn't allow changing the admin address through the edit, that will be its own transaction to prevent mistakes when making edits to the config.

I introduced a length limit to the memo to prevent potential abuse. 3000 characters feels like it should be enough for custom applications or whatever fun messages people want to attach to their proxy.

I took inspiration from the x/staking module on how they deal with optional fields. I realised making them nullable doesn't do much since we'll just get an empty string which could also be the value the user tried to pass. Now we check for some constants and rely on the application creating the transaction to supply the correct values in case there are no changes required.

The fee update code feels a little convoluted, but I'm not sure how to express it more clearly with just code. Any suggestions are welcome. :)

## Testing

Added tests to msg_server_test that deal with various edit scenarios.

## Related PRs and Issues

Part-of: #316


x/data-proxy/keeper/keeper.go
```diff
@@ -20,15 +20,16 @@ type Keeper struct {
 
 	Schema           collections.Schema
 	DataProxyConfigs collections.Map[[]byte, types.ProxyConfig]
-	// TODO queue
-	Params collections.Item[types.Params]
+	FeeUpdates       collections.KeySet[collections.Pair[int64, []byte]]
```
hacheigriega, line 23: I think `FeeUpdateQueue` or something would be more suitable

x/data-proxy/keeper/keeper.go
```diff
@@ -58,3 +59,24 @@ func (k Keeper) GetDataProxyConfig(ctx context.Context, pubKey string) (result t
 
 	return config, nil
 }
+
+func (k Keeper) GetFeeUpdates(ctx context.Context, activationHeight int64) ([][]byte, error) {
```
hacheigriega, line 63: `GetFeeUpdatePubKeys`?

x/data-proxy/keeper/msg_server.go
```diff
@@ -67,23 +70,118 @@ func (m msgServer) RegisterDataProxy(goCtx context.Context, msg *types.MsgRegist
...
+			return nil, err
+		}
+
+		return &types.MsgEditDataProxyResponse{}, nil
+	}
+
+	return m.ProcessProxyFeeUpdate(ctx, pubKeyBytes, &proxyConfig, msg)
+}
+
+func (m msgServer) ProcessProxyFeeUpdate(ctx sdk.Context, pubKeyBytes []byte, proxyConfig *types.ProxyConfig, msg *types.MsgEditDataProxy) (*types.MsgEditDataProxyResponse, error) {
```
hacheigriega, line 136: I think this should be turned into a keeper method from a msg server method because it is not one of the msg server interface methods.

x/data-proxy/types/tx.go
```diff
@@ -20,11 +20,38 @@ func (m *MsgRegisterDataProxy) Validate() error {
...
+	}
+
+	if hasNewPayoutAddress {
+		if _, err := sdk.AccAddressFromBech32(m.NewPayoutAddress); err != nil {
+			return ErrInvalidAddress.Wrap(err.Error())
+		}
+	}
+
+	if hasNewFee {
+		// TODO validate fee is in native denom
```
hacheigriega, line 51: Unfortunately I think this check will have to be performed in msg server because there is no keeper access here..

x/data-proxy/keeper/msg_server.go
```diff
@@ -67,23 +70,118 @@ func (m msgServer) RegisterDataProxy(goCtx context.Context, msg *types.MsgRegist
...
+
+	// Determine update height
+	updateHeight := ctx.BlockHeight() + int64(updateDelay)
+	feeUpdate := &types.FeeUpdate{
+		NewFee:       *msg.NewFee,
+		UpdateHeight: updateHeight,
+	}
+
+	// Retain previous pending update so it can be removed
+	previousPendingUpdate := proxyConfig.FeeUpdate
```
hacheigriega, line 160: Why don't we just check the value of `proxyConfig.FeeUpdate` and remove if applicable here?

x/data-proxy/keeper/msg_server.go
```diff
@@ -67,23 +70,118 @@ func (m msgServer) RegisterDataProxy(goCtx context.Context, msg *types.MsgRegist
...
+		}
+	}
+
+	err = m.DataProxyConfigs.Set(ctx, pubKeyBytes, *proxyConfig)
+	if err != nil {
+		return nil, err
+	}
+
+	return &types.MsgEditDataProxyResponse{
+		UpdateHeight: updateHeight,
```
hacheigriega, line 183: I think `FeeUpdateHeight` for the field name would be clearer.

x/data-proxy/keeper/msg_server.go
Thomasvdam: Great question, no real reason.

x/data-proxy/keeper/keeper.go
```diff
@@ -58,3 +60,70 @@ func (k Keeper) GetDataProxyConfig(ctx context.Context, pubKey string) (result t
...
+	}
+
+	for _, k := range keys {
+		pubkeys = append(pubkeys, k.K2())
+	}
+
+	return pubkeys, nil
+}
+
+func (k Keeper) ProcessProxyFeeUpdate(ctx sdk.Context, pubKeyBytes []byte, proxyConfig *types.ProxyConfig, msg *types.MsgEditDataProxy) (int64, error) {
```
hacheigriega, line 85: Since this method's logic is a little dependent on `msgServer.EditDataProxy`, we can make it unexported? Also, it can take as argument a `types.FeeUpdate` struct or just `msg.FeeUpdateDelay` & `msg.NewFee`.

hacheigriega (approved):

PR #336, opened 2024-08-21
Additional validation and improved errors

Thomasvdam:
## Motivation

Make errors more in line with the changes introduced in #333 and adds additional validation on TXs.

## Explanation of Changes

We decided to check the denom through the appparam const as it seems that changing the denom in the staking keeper wouldn't work anyway.

## Testing

Add unit tests to verify that param changes don't affect pending updates.

## Related PRs and Issues

Part-of: #316


hacheigriega (approved):
Great work.. 
Found that we're still creating an error arbitrarily here https://github.com/sedaprotocol/seda-chain/blob/36761462f7e2c11c9a92694af023afbd7eb3f9d9/x/data-proxy/keeper/msg_server.go#L230

gluax (approved):

PR #337, opened 2024-08-21
Data proxy cleanup

Thomasvdam:
## Motivation

Small cleanup removing some nullable proto annotations as they didn't seem to add any value but made the code a little more difficult to deal with.

As discussed we prefer not to interact with the stores directly when it's not from the immediate Keeper context. This means most get/set/has operations need their own wrapper so the boundaries that expose the Keeper don't need to have knowledge of the actual store implementation.

## Explanation of Changes

I didn't find a way to cleanly prevent the boundaries (abci.go, msg_server.go, grpc_query.go) from accessing the stores directly, but at least most operations now have their own API and the boundary doesn't need to know anything about how the store works internally.

## Testing

Verify tests still pass

## Related PRs and Issues

Should be the last part of #316


x/data-proxy/keeper/keeper.go
```diff
@@ -49,72 +48,87 @@ func (k Keeper) GetAuthority() string {
...
+
+	itr, err := k.feeUpdateQueue.Iterate(ctx, rng)
+	if err != nil {
+		return nil, err
+	}
+
+	keys, err := itr.Keys()
+	if err != nil {
+		return nil, err
+	}
```
gluax, line 119: Can `Go` steal `rust`'s `?` operator already 😢

gluax (approved):

x/data-proxy/keeper/params.go
```diff
@@ -0,0 +1,16 @@
+package keeper
+
+import (
+	sdk "github.com/cosmos/cosmos-sdk/types"
+
+	"github.com/sedaprotocol/seda-chain/x/data-proxy/types"
+)
+
+func (k Keeper) StoreParams(ctx sdk.Context, params types.Params) error {
```
hacheigriega, line 9: `SetParams` just to follow SDK convention?

x/data-proxy/keeper/keeper.go
```diff
@@ -49,72 +48,87 @@ func (k Keeper) GetAuthority() string {
...
-func (k Keeper) GetDataProxyConfig(ctx context.Context, pubKey string) (result types.ProxyConfig, err error) {
-	pubKeyBytes, err := hex.DecodeString(pubKey)
-	if err != nil {
-		return types.ProxyConfig{}, err
-	}
+func (k Keeper) HasDataProxy(ctx sdk.Context, pubKey []byte) (bool, error) {
+	return k.dataProxyConfigs.Has(ctx, pubKey)
+}
+
+func (k Keeper) StoreDataProxyConfig(ctx context.Context, pubKey []byte, proxyConfig types.ProxyConfig) error {
```
hacheigriega, line 55: `SetDataProxyConfig`

x/data-proxy/keeper/keeper.go
```diff
@@ -49,72 +48,87 @@ func (k Keeper) GetAuthority() string {
...
-
-	keys, err := itr.Keys()
-	if err != nil {
-		return nil, err
-	}
-
-	for _, k := range keys {
-		pubkeys = append(pubkeys, k.K2())
-	}
+func (k Keeper) StoreFeeUpdate(ctx sdk.Context, height int64, pubKey []byte) error {
```
hacheigriega, line 68: `SetFeeUpdate` just to follow SDK convention

hacheigriega (approved):

PR #342, opened 2024-08-26
feat: Pubkey module

hacheigriega:


Thomasvdam:
I guess it's not entirely clear for me how this module interfaces with modules that need access to the private keys, or how the dependencies are managed.

Maybe it helps to illustrate the following scenarios:
1. We create a new module that needs to make use of the PKR, how does this new key get added?
2. A module needs to have something signed, how does this happen?

It would be nice if we could leave all the details of how the keys are managed/stored/etc in the PKR module. I don't think it's possible to express this purely in terms of method/function calls and we'll have to keep a manual file somewhere that links a consumer module identifier to the kind of key it needs.

- New module needs a keypair: add an entry to a list that consists of an ID (the index, should be unique) and the key implementation (how to generate, how to sign?)
- On add-keys command the CLI iterates over the list, generates keys at the index of that module, publishes the new public keys in the TX. Private keys get stored in 1 file.
- New module needs something signed, calls a method on the PKR module to sign data with the key under ID X, PKR returns signed bytes.

Pretty sure I'm simplifying things too much, but I feel like this should be possible.

hacheigriega:
1. My thinking was that when we release a new binary that expects some new key to be registered, we would also update the CLI and have the validators use the updated CLI to generate a new set of keys and register their pubkeys.
2. My thinking at the moment is that signing and verifying should be separated. Signing should be done using utility or context, whereas verifying should be supported by the pubkey module, which would expose the method through expected keepers. I think this separation makes sense because signing deals with the key file and is relevant only to validators. Would love to discuss more if you have doubts, but this is my current thinking.

hacheigriega:
I will remove unused code like VRF key or CLI endpoint for creating validator with VRF in a separate PR tomorrow

hacheigriega:
The key interface and the linting error are addressed in PR #365.

app/utils/vrf_key.go
```diff
@@ -225,23 +202,39 @@ func LoadVRFKey(keyFilePath string) (*VRFKey, error) {
...
-		return nil, fmt.Errorf("could not create directory %q: %w", filepath.Dir(pvKeyFile), err)
-	}
+// LoadOrGenVRFKey initializes a VRF key and returns its public key.
+// If loadPath is specified, it loads the VRF key file at the specified
+// path. Otherwise, it generates a new VRF key, whose entropy is randomly
+// generated or obtained from the mnemonic, if provided.
+func LoadOrGenVRFKey(config *cfg.Config, loadPath string) (vrfPubKey sdkcrypto.PubKey, err error) {
+	var vrfKey *VRFKey
+	if loadPath != "" {
+		vrfKey, err = LoadVRFKey(loadPath)
```
gluax, line 212: Is this standard for cosmos? I'd say it's a risk to store a private key as plain text in a file. So we should additionally have the option to do it as an ENV var.

x/pubkey/client/cli/tx.go
```diff
@@ -0,0 +1,202 @@
...
+			Index:  key.Index,
+			PubKey: pkAny,
+		}
+	}
+	return result, nil
+}
+
+// saveSEDAKeys saves a given list of IndexedPrivKey in the directory
+// at dirPath.
+func saveSEDAKeys(keys []IndexedPrivKey, dirPath string) error {
```
gluax, line 140: These are private keys right? This is fine for devnet. Cosmos offers the file backend for a similar case, but they do specify you should NOT use file storage for an official deployment.

This security stuff could also done in another pr. Also maybe we could leverage the existing system cosmos has to use a different storage backed.

x/pubkey/client/cli/tx_test.go
```diff
@@ -0,0 +1,140 @@
...
+	testutilmod "github.com/cosmos/cosmos-sdk/types/module/testutil"
+
+	"github.com/sedaprotocol/seda-chain/app/params"
+	"github.com/sedaprotocol/seda-chain/x/pubkey/client/cli"
+)
+
+type CLITestSuite struct {
+	suite.Suite
+
+	kr          keyring.Keyring
```
gluax, line 35: Oh, I do see this here, can we just use the keyring to store these private keys? In a dev environment that works cause they allow you to use the file backend.

gluax (commented):
My comments may be me not understanding the cosmos SDK too well. But it's a security issue if we store private info in plain text in a file.

app/utils/vrf_key.go
hacheigriega: created an issue based on what we discussed: #368

x/pubkey/client/cli/tx.go
hacheigriega: created an issue #368

x/pubkey/client/cli/tx_test.go
hacheigriega: Keyring is unsuitable for this purpose because we need repeated access to the key without the key owner's approval every time.

Thomasvdam (approved):

PR #343, opened 2024-08-26
chore: Update dependencies

hacheigriega:
- Bumping SEDA Tally VM version from v1.0.1 to v1.1.0.
- Bumping Cosmos SDK version from v0.50.6 to v0.50.9. 
- Bumping Comet BFT version from v0.38.6 to v0.38.11.
- Bumping wasmvm version from v1.5.0 to v1.5.4.
- Could not update wasmd because any version newer than our current version requires wasmvm >= v2.0.0.


Thomasvdam (approved):
Is the failed simulation a problem? Seems odd that all validators unbond within 86 blocks

hacheigriega:
We've been ignoring it as it's likely an SDK issue. I see that there was a related merge recently https://github.com/cosmos/cosmos-sdk/pull/18196 
I'll create an issue to look into it more deeply

Issue #317, opened 2024-08-12
✨ Data Proxy in Tally Consensus


Thomasvdam:
## ✨ Feature

As part of the data proxy feature of the SEDA network the overlay nodes will start submitting the public keys of data proxy nodes that were queried during WASM execution. The overlay nodes should be in consensus on what data sources were queried in order to determine execution validity and fee payout.

## Motivation

Prevent overlay nodes from submitting public keys that were not actually queried in an attempt to receive additional rewards.

## Implementation

Consensus will be computed with the tuple of `(exit_code, public_keys)` and a `mode` filter.

## Related Issues

Blocked by https://github.com/sedaprotocol/seda-chain-contracts/issues/201

PR #344, opened 2024-08-26
feat: Add basic consensus check to support data proxy tally

hacheigriega:
## Explanation of Changes

Until now, consensus was determined solely based on the reveal data at a given data path (Except when None filter is applied, in which case there is no check for consensus at all). Following @Thomasvdam 's proposal, this PR adds an additional "basic consensus check," which checks for consensus on the tuple `(exit_code, pub_keys)` before filter is applied. If less than 2/3 of reveals have the same tuple, a newly added error type `ErrNoBasicConsensus` is returned so that tally VM is not executed.

Note that when checking for basic consensus, the specific exit code values do not matter. Whether an exit code is a success or not is checked in the function `parseReveals()`, which is run by all filter types except for None filter. If more than 1/3 of the reveals have non-zero exit codes, the error `ErrCorruptReveals` is returned.

For example, 
```
exit_code, data
1,A
1,A
0,A
0,A
```

results in a "no basic consensus" error, while

```
exit_code, data
1,A
1,A
1,A
0,A
```
results in a corruption error.

Also note that in case of "no basic consensus" error, an all-zero outlier list is returned and tally VM is not executed, unlike in case of the general consensus error.

Updated filter specs [here](https://www.notion.so/sedaprotocol/Tallying-9388fd590d1c4b659abb721b57909feb?pvs=4#359ad569a9e441ca9e031be61d9a49a2)

## Testing

Additional test cases were added to `TestFilter`.

## Related PRs and Issues

Closes: #317


x/tally/types/errors.go
```diff
@@ -7,8 +7,9 @@ var (
 	ErrFilterInputTooShort = errors.Register("tally", 3, "filter input length too short")
 	ErrInvalidPathLen      = errors.Register("tally", 4, "invalid JSON path length")
 	ErrEmptyReveals        = errors.Register("tally", 5, "no reveals given")
-	ErrCorruptReveals      = errors.Register("tally", 6, "more than 1/3 of the reveals are corrupted")
-	ErrNoConsensus         = errors.Register("tally", 7, "1/3 or more of the reveals are not in consensus range")
-	ErrInvalidNumberType   = errors.Register("tally", 8, "invalid number type specified")
-	ErrFilterUnexpected    = errors.Register("tally", 9, "unexpected error occurred in filter")
+	ErrCorruptReveals      = errors.Register("tally", 6, "> 1/3 of reveals are corrupted")
+	ErrNoConsensus         = errors.Register("tally", 7, "> 1/3 of reveals do not agree on reveal data")
+	ErrNoBasicConsensus    = errors.Register("tally", 8, "> 1/3 of reveals do not agree on (exit_code, proxy_pub_keys)")
```
Thomasvdam, line 12: Not a huge fan of the name but I don't have a better alternative 😅

x/tally/keeper/filter.go
```diff
@@ -15,12 +16,29 @@ const (
...
 	}
 
+	// Determine basic consensus on tuple of (exit_code, proxy_pub_keys)
+	var maxFreq int
+	freq := make(map[string]int, len(reveals))
+	for _, reveal := range reveals {
+		var success bool
+		if reveal.ExitCode == 0 {
+			success = true
+		}
```
Thomasvdam, line 33: Can you do `success := reveal.ExitCode == 0` in Golang or is considered not done?

Thomasvdam (commented):
I think it would be nice if we can emit the 'agreed' upon public keys in the tally event. That will allow us to display it in the explorer and should help with debugging :)

x/tally/keeper/filter.go
hacheigriega: Good points. That works but now Im thinking whether we should just use the exit code value rather than success boolean here. Wdyt?

Thomasvdam: I think the result will be the same so maybe that makes more sense... Although thinking about it maybe we should do the true/false thing as we also want to be able to short circuit for the None filter?

x/tally/types/filters.go
```diff
@@ -24,8 +24,23 @@ func NewFilterNone(_ []byte) (FilterNone, error) {
 	return FilterNone{}, nil
 }
 
-// FilterNone declares all reveals as non-outliers.
+// FilterNone declares all reveals as non-outliers, unless there
```
Thomasvdam, line 27: Unless there…..

Thomasvdam (approved):
The only thing I'm unsure about it whether we should emit the data-proxy pubkeys when there isn't 2/3rds with a successful execution. But this won't really matter until we do incentives so I think we should leave it like this and once we decide how the rewards are done for all the different scenarios we can adjust the event accordingly. :)

hacheigriega:
> The only thing I'm unsure about it whether we should emit the data-proxy pubkeys when there isn't 2/3rds with a successful execution. But this won't really matter until we do incentives so I think we should leave it like this and once we decide how the rewards are done for all the different scenarios we can adjust the event accordingly. :)

Noted. Thanks for the review!

PR #349, opened 2024-08-27
feat: Pass additional env variables to tally VM

### LINKED ISSUE #318: ✨ Additional Tally VM env variables
## ✨ Feature

Right now we only pass the `VM_MODE` and `CONSENSUS` environment variables to the tally VM. We should provide the VM with additional information on the data request.

Also see https://www.notion.so/sedaprotocol/Tallying-9388fd590d1c4b659abb721b57909feb?pvs=4#39cd1e38c8784beb9fe952dc69dfd567 for more information.

## Motivation

This could enable new usecases in the tally phase and since we already have the data the effort of forwarding it should be relatively low.

## Implementation

Besides the given arguments, the VM also passes a few extra environment variables for the Tally binary to use (not limited to):

- DR_ID
- DR_INPUT
- BINARY_ID
- DR_REPLICATION_FACTOR
- DR_GAS_PRICE
- DR_GAS_LIMIT
- DR_MEMO
- DR_PAYBACK_ADDRESS

The more information a tally VM can have, the better.

- [x] Have an oracle program that prints out all env variables in tally execution. Ask me on Slack

hacheigriega:
## Explanation of Changes

When executing tally, we now pass additional information about the data request as environment variables. 

## Testing

Tested using a sample tally wasm that prints all the environment variables it accepts. 

## Related PRs and Issues

Closes #318


Thomasvdam (approved):
@FranklinWaller Do you think it makes sense to include the height at which the DR was issued?

x/tally/keeper/abci.go
```diff
@@ -200,8 +200,16 @@ func (k Keeper) FilterAndTally(ctx sdk.Context, req types.Request) (tallyvm.VmRe
...
+		"VM_MODE":               "tally",
+		"CONSENSUS":             fmt.Sprintf("%v", consensus),
+		"DR_ID":                 req.ID,
+		"DR_INPUT":              req.DrInputs,
+		"BINARY_ID":             req.DrBinaryID,
+		"DR_REPLICATION_FACTOR": fmt.Sprintf("%v", req.ReplicationFactor),
+		"DR_GAS_PRICE":          req.GasPrice,
+		"DR_GAS_LIMIT":          req.GasLimit,
+		"DR_MEMO":               req.Memo,
+		"DR_PAYBACK_ADDRESS":    req.PaybackAddress,
```
FranklinWaller, line 212: I can't really tell from this, but are these hex encoded instead of base64?

FranklinWaller:
> @FranklinWaller Do you think it makes sense to include the height at which the DR was issued?

If we have that information we should give it

x/tally/keeper/abci.go
hacheigriega: base64. In fact, the Sudo call to the contract fails if it's in hex or something other than base64 encoding

hacheigriega: Ah Im just saying that the contract assumes it to be a base64 string. So if VM wants it to be hex-encoded, we will have to convert it on chain.

hacheigriega:
OK Converted base64-encoded payback address to hex before passing it to tally VM as an environment variable and also added block height as an env var.

Thomasvdam (approved):

PR #354, opened 2024-09-04
feat(batching): Batching module with batch construction in end blocker

### LINKED ISSUE #351: ✨ Create `x/batching` and construct batch in end blocker
## ✨ Feature

As a first step towards initial implementation of the batching module, add an end blocker logic for constructing batches. This requires constructing two "ordered" merkle trees where the left child node's hash in hex is always alphanumerically smaller than the right child node's. 

A batch type would look something like this:
```go
type Batch struct {
	batchNumber    int
	blockHeight    int
	dataResultRoot string
	validatorRoot  string
	votes          []Vote
}

type Vote struct {
	validatorAddr string // sedavaloper address
	votingPower   int
	signature     []Signature // signatures for different schemes
}

type Signature struct {
	scheme      string
	signature   string
	publicKey   string // registered in pubkey module
	merkleProof string
}
```

And the states kept by the batching module would be:
```
0x00 -> current_batch_number
0x01 | batch_number -> Batch
```

hacheigriega:
## Batching Module

This PR adds the initial module boilerplate for `x/batching` along with its end blocker, where batches are constructed and stored. The end blocker constructs two merkle trees, data result tree and validator tree, at every block and persists their roots and entries (unhashed leaves) in a batch. Please note that the data result tree implementation will be completed in another PR as it will involve some store refactor.

Please see the specs page for additional details.

## Merkle Tree

~~Our merkle tree implementation largely follows [RFC-6962](https://datatracker.ietf.org/doc/html/rfc6962#section-2.1) but with some modifications:~~

Our merkle tree implementation follows the OpenZeppelin standards as described [here](https://github.com/OpenZeppelin/merkle-tree#simple-merkle-trees).
- It uses Keccak-256 hashing algorithm.
- Leaves, which are single-hashed entries, are sorted.
- Each hash pair is sorted.

## Testing

The integration test randomly generates 100 validators with secp256k1 public keys, randomly generates a proof for each of the validator entries, and then verifies them against the computed root. 

## Related PRs and Issues

Related Issue #351


x/batching/keeper/abci.go
```diff
@@ -0,0 +1,85 @@
...
+	// TODO: Deal with offset and limits. (#313)
+	// queryRes, err := k.wasmViewKeeper.QuerySmart(ctx, coreContract, []byte(`{"get_data_requests_by_status":{"status": "tallied", "offset": 0, "limit": 100}}`))
+	// if err != nil {
+	// 	return err
+	// }
+	// if string(queryRes) == "[]" {
+	// 	return nil
+	// }
+
+	// Leaves are SHA-256 hashes of JSON-serialized data result structs.
```
Thomasvdam, line 60: Aren't the data result IDs already hashes of the data result?

hacheigriega: We confirmed that they are, so we will use the data result ID as the tree's leaves.

x/batching/keeper/abci.go
```diff
@@ -52,69 +52,68 @@ func (k Keeper) ConstructBatch(ctx sdk.Context) (types.Batch, error) {
...
 	return types.Batch{
-		BatchNumber:    curBatchNum,
-		BlockHeight:    ctx.BlockHeight(),
-		DataResultRoot: dataRootHex,
-		ValidatorRoot:  valRootHex,
-		BlockTime:      ctx.BlockTime(),
+		BatchNumber:      curBatchNum,
+		BlockHeight:      ctx.BlockHeight(),
+		DataResultRoot:   dataRootHex,
+		ValidatorRoot:    valRootHex,
```
Thomasvdam, line 68: Maybe too soon but we ultimately want a single root hash right? With the previous batch root as 1 direct child

hacheigriega: I don't think we reached an agreement on what exactly to have the validators sign. My understanding was that we want to have them sign the serialized batches, which would also include batch number, block height, etc. Having a single root would facilitate proving, though.

cmd/sedad/utils/merkle.go
```diff
@@ -0,0 +1,107 @@
...
+// RootFromEntries computes the root of the Merkle Hash Tree whose
+// leaves are the entries given in byte slices in the provided order.
+// It largely follows RFC-6962 with some modifications:
+//   - It uses SHA3-256 hashing algorithm.
+//   - Leaves are double-hashed for second preimage resistance. No hashing
+//     prefixes are used.
+//   - Each hash pair is sorted to make proofs more succinct.
+//   - "Super root" computation is supported.
+func RootFromEntries(entries [][]byte) []byte {
+	return rootFromEntries(sha3.New256(), entries)
```
mariocao, line 20: I think that this hashing algorithm is not the same as `keccak256`, which is widely used in blockchains.

Usually, keccak256 is the algorithm that won the NIST competition, but then during the SHA3 standarization they did minor changes that ended up to, what I think, it is the New256. Best is to check with some test vectors! :D

x/batching/keeper/abci.go
mariocao: We should aim to single root to be signed and not multiple. We should quickly agree on a tree structure. Let's discuss this. 💬

cmd/sedad/utils/merkle.go
hacheigriega: Ah, didn't realize this. I see that geth uses legacy keccak256. Updated the hash function and added some unit tests.

hacheigriega:
I will address the lint errors in a separate PR. Some of those come from outdated code like VRF that has to be removed.

x/pubkey/client/cli/tx.go
Thomasvdam, line 1: The linter has an opinion on some lines in this file.

x/batching/types/params.go
```diff
@@ -0,0 +1,17 @@
...
+// DefaultParams returns default batching module parameters.
+func DefaultParams() Params {
+	return Params{
+		ValidatorSetTrimPercent: DefaultValSetTrimPercent,
+	}
+}
+
+// Validate validates the batching module parameters.
+func (p *Params) Validate() error {
+	return nil
```
Thomasvdam, line 16: Do we want to check that the trimPercent does not exceed 100%?

x/batching/keeper/abci.go
Thomasvdam: Didn't we agree to have a single root?

cmd/sedad/utils/merkle.go
```diff
@@ -0,0 +1,107 @@
...
+	"hash"
+	"math/bits"
+
+	"golang.org/x/crypto/sha3"
+)
+
+// RootFromEntries computes the root of the Merkle Hash Tree whose
+// leaves are the entries given in byte slices in the provided order.
+// It largely follows RFC-6962 with some modifications:
+//   - It uses SHA3-256 hashing algorithm.
```
Thomasvdam, line 14: Is this still true?

cmd/sedad/utils/merkle.go
```diff
@@ -0,0 +1,107 @@
...
+//   - Leaves are double-hashed for second preimage resistance. No hashing
```
Thomasvdam, line 15: Same question here?

cmd/sedad/utils/merkle.go
hacheigriega: ah right. changed "SHA3-256" to "Keccak-256"

cmd/sedad/utils/merkle.go
hacheigriega: We do still double hash the leaves if that is what you're referring to?

x/batching/types/params.go
hacheigriega: good point - fixed

x/pubkey/client/cli/tx.go
hacheigriega: fixed one of those and the others will be addressed in #365

cmd/sedad/utils/merkle.go
Thomasvdam: Yeah since when we discussed it I thought we concluded that the preimage attack vector didn’t really apply to us + DR result IDs are already a hash of the dr result. No opinion, just curious to what we landed on ultimately

hacheigriega: Yes makes sense. I don't think we want to finalize the specs before we try out proving from the Solidity contract side, but for now:

- Leaves are sorted
- Each hash pair is sorted to make proofs more succinct.
- No hashing prefixes or double hashing leaves

x/batching/types/errors.go
```diff
@@ -0,0 +1,7 @@
+package types
+
+import "cosmossdk.io/errors"
+
+var (
+	ErrInvalidValidatorSetTrimPercent = errors.Register(ModuleName, 2, "validator set trim percent must be between 0 and 100")
```
Thomasvdam, line 6: Do we want a custom error for this or use the standard invalid value one?

x/batching/keeper/abci.go
```diff
@@ -0,0 +1,149 @@
...
+				return false
+			}
+			panic(err)
+		}
+
+		pkBytes := pubKey.Bytes()
+		entry := make([]byte, len(pkBytes)+8)
+
+		copy(entry[:len(pkBytes)], pkBytes)
+		binary.BigEndian.PutUint64(entry[len(pkBytes):], uint64(power))
```
Thomasvdam, line 138: Weren't we going to use a uint16 or uint32 for a percentage based representation of the voting power?

cmd/sedad/utils/merkle_proof.go
```diff
@@ -0,0 +1,91 @@
...
+	})
+
+	// Build the tree.
+	tree := buildTree(sha, leaves)
+
+	// Find the index of the entry in the tree.
+	treeIndex := -1
+	want := leafHash(sha, entries[index])
+	for i := range tree {
+		if bytes.Equal(want, tree[i]) == true {
```
gluax, line 38: No need to compare to a boolean here.

x/batching/types/errors.go
gluax: I think I like the more specific ones more as it clarifies what the problem is?

gluax (approved):

x/batching/keeper/abci.go
hacheigriega: You're right. And I updated the code, but I need to think about possible overflow when casting to uint32..

gluax: Yeah, worrying about the overflow would be good. You could `mod` by `u32::MAX` or compare it to u32 max. Though golang might have some way to deal with already built-in, and I'm not familiar enough with it. On overflows, we could either error or cap it.

x/batching/types/errors.go
Thomasvdam: You can still add your own message to the standard invalid value error, so it won’t be complete guesswork as to what the invalid value was :)

x/batching/keeper/abci.go
```diff
@@ -0,0 +1,171 @@
...
+			if errors.Is(err, collections.ErrNotFound) {
+				return false
+			}
+			panic(err)
+		}
+		validators = append(validators, validatorData{
+			secp256k1PubKey: pubKey,
+			power:           math.NewInt(power).MulRaw(1e8),
+		})
+		totalPower = totalPower.AddRaw(power)
```
Thomasvdam, line 145: There is total power on the staking keeper right?

hacheigriega: updated the code

x/batching/keeper/abci.go
```diff
@@ -0,0 +1,161 @@
...
+// ConstructDataResultTree constructs a data result tree based on the
+// batching-ready data results returned from the core contract and
+// returns its leaves and its hex-encoded tree root.
+func (k Keeper) ConstructDataResultTree(ctx sdk.Context) ([][]byte, string, error) {
+	coreContract, err := k.wasmStorageKeeper.GetCoreContractAddr(ctx)
+	if err != nil {
+		return nil, "", err
+	}
+	// TODO: Deal with offset and limits. (#313)
+	queryRes, err := k.wasmViewKeeper.QuerySmart(ctx, coreContract, []byte(`{"get_data_results_by_status":{"status": "tallied", "offset": 0, "limit": 100}}`))
```
gluax, line 85: wonder if would be safer/better to have this as a struct instead of raw JSON

gluax (approved):

x/batching/keeper/abci.go
```diff
@@ -0,0 +1,161 @@
...
+		secp256k1PubKey, err := k.pubKeyKeeper.GetValidatorKeyAtIndex(ctx, valAddr, utils.SEDAKeysIndexSecp256k1)
+		if err != nil {
+			if errors.Is(err, collections.ErrNotFound) {
+				return false
+			}
+			panic(err)
+		}
+
+		// Compute validator voting power percentage.
+		powerPercent := math.NewInt(power).MulRaw(1e8).Quo(totalPower).Uint64()
```
Thomasvdam, line 141: Maybe good to have a TODO for the validator set trimming?

x/batching/keeper/abci.go
hacheigriega: It's mostly just to avoid marshaling it every time we use it..

Issue #362, opened 2024-09-23
🐛 SQS connection resets do not get retried


Thomasvdam:
## 🐛 Bug Report

Every once in a while the indexing plugin receives a connection rest while attempting to send a message to SQS, similar to the S3 issue (#273). This halts the indexing process as there are some cases where an error in the indexer warrants a reset to the previous block. By default the retry logic from the AWS SDK does not retry on connection resets as it can't know whether the request that was made is idempotent or not. Since we know that the upload action is idempotent we can safely configure the retry SDK to retry connection reset errors.

## Expected Behavior

The indexing plugin should retry on connection resets.

## Your Environment

N.A.

PR #363, opened 2024-09-23
Add connection reset retry to SQS

Thomasvdam:
## Motivation

See #362 

## Explanation of Changes

Make it easier to apply the retry logic to different configs, and the config to the SQS client.

Also disable the staking messages in the plugin. Before we kept an allowlist of message types that didn't include anything from the staking module, but as it'll probably be a while before we add support for those I figure it's easier to just disable the functionality at the plugin level.

## Testing

Confirm messages still get sent locally. Difficult to trigger a connection reset error locally though, but worst case it doesn't retry which is the same behaviour we have right now.

## Related PRs and Issues

Closes: #362


mariocao (approved):
lgtm

Issue #314, opened 2024-07-30
✨ Re-enable `x/randomness` with PKR support


hacheigriega:
## ✨ Feature

Re-enable `x/randomness` that is currently not wired to the application. After wiring the module, it is also important to make sure that key generation process is clearly defined both for existing validators with a consensus key and new validators without any keys. The module should also use the newly-implemented module `x/pkr` for all public key related operations.

PR #365, opened 2024-09-23
refactor: Remove VRF and add SEDA keys interface

### LINKED ISSUE #364: ♻️SEDA key interface
## ♻️ Refactor

The SEDA key utilities should expose an interface, which various modules and ABCI methods will then use to support signing given input messages, signing transactions, or verifying signatures. This interface should encapsulate access to the key file (for signing only) and serve as a source of truth regarding the indices and the corresponding cryptographic curves as defined in the SEDA Protocol.

```
type sedaKey interface {
    Sign(input []byte, index uint32) (proof []byte, err error)
    Verify(publicKey, input, proof []byte) (err error)
}
```

The utility should have a load function that takes a key file path as an input and returns an object that satisfies this interface.

This issue also calls for 

- Defining constant values regarding indices and the corresponding curves to be used by the pubkey module.
- Removal of currently unused code like VRF signer, randomness module, and the custom staking module wrapper's create-validator-with-vrf tx.
- Address linting errors.

hacheigriega:
## Explanation of Changes

This PR replaces the existing (but unused) VRF signer interface with a new interface `sedaSigner`. The application can call `LoadSEDASigner(loadPath string)` to attain an object that satisfies the interface. This interface can then be used by various ABCI methods to provide signing services for the validators.

The SEDA keys utility will also serve as a single source of truth regarding the indices and their corresponding cryptographic curves. 

This PR also removes the following unused code:
- Remove `x/randomness` for now (Related issue #314).
- Remove the custom staking module's `create-validator-with-vrf` and the corresponding gentx logic.

## Related PRs and Issues

Related #314
Closes #364


Thomasvdam (commented):
It would be nice if we can make the key index a known constant. So the batching module can do something like 
```golang
import (
  "github.com/sedaprotocol/seda-chain/app/utils"
)

// Get a SEDASigner
sedaSigner.Sign(input, utils.Secp256k1_index)
```
I do realise that this is basically a map with extra steps, so maybe that's worth considering?

hacheigriega:
> I do realise that this is basically a map with extra steps, so maybe that's worth considering?

Yeah great point. I did think about it before, but the general phobia of maps held me back. It shouldn’t be an issue here since the map iterations will be done off-chain. I’d appreciate your second look!

Thomasvdam (approved):

PR #370, opened 2024-10-01
feat(batching): vote extensions for signing batches and verifying signatures

### LINKED ISSUE #352: ✨ Batching signing and signature verification
## ✨ Feature

The second phase of the batching implementation involves signing the batch created in the end blocker. We will utilize the ABCI 2.0 methods `ExtendVote` to add batch signatures to the pre-commit votes and `VerifyVoteExtension` to verify the signatures before accepting the votes. Since batch is created at the very last stage of the consensus process (unless we modify the SDK code), the batch signing and signature verification will take place in the next block height.

hacheigriega:
### Explanation of Changes

This PR implements ABCI++ methods `ExtendVote` and `VerifyVote` for signing and verifying batch votes, respectively. The node only enables the `ExtendVote` method when a valid SEDA key file is successfully loaded. If it is not successfully loaded, only `VerifyVote` method will be activated. This means that the validator's votes will be rejected by the other validators once the block height reaches the parameter `vote_extensions_enable_height`.

Since `VerifyVote` request contains the consensus key address, the pubkey module has been updated to store the SEDA keys' public keys under the consensus key address, instead of the validator operator address.

### Testing

A local network of two validators can be spawned by running `./scripts/local_multi_setup.sh`. This script starts the two nodes using tmux with the `vote_extensions_enable_height` configured to 15. The two validators then generate the SEDA keys and add them on chain through the pubkey module transaction. Finally, the two validator nodes are restarted to activate the SEDA key signers.

Closes #352


app/abci/vote_extension.go
```diff
@@ -0,0 +1,74 @@
...
+func (h *VoteExtensionHandler) ExtendVoteHandler() sdk.ExtendVoteHandler {
+	return func(ctx sdk.Context, _ *cometabci.RequestExtendVote) (*cometabci.ResponseExtendVote, error) {
+		h.logger.Debug("start extend vote handler")
+		batch, err := h.batchingKeeper.GetCurrentBatch(ctx)
+		if err != nil {
+			return nil, err
+		}
+		if batch.BlockHeight != ctx.BlockHeight()-1 {
+			return nil, ErrNoBatchForCurrentHeight
+		}
```
Thomasvdam, line 38: I think it's a little clearer if we make it explicit that we want the batch of the previous height. So something like `batchingKeeper.GetBatchForHeight(ctx, ctx.BlockHeight()-1)`.

I'm not sold on that approach btw, I just want to avoid confusion where we say current, but then verify if it's the currentHeight-1. I'm also fine with something like `GetPreviousBlockBatch` on the Keeper and have that take care of the height shenanigans.

Thomasvdam (commented):
I'm not sold on using the consensus key in the Pubkey registry. I get that it saves a lookup during the vote extension verification but given that the next version of the Cosmos SDK introduces consensus key rotation I think we're setting ourselves up for difficulties.

x/batching/keeper/keeper.go
```diff
@@ -140,6 +140,18 @@ func (k Keeper) GetBatch(ctx context.Context, batchNum uint64) (types.Batch, err
...
 
+func (k Keeper) GetCurrentBatch(ctx context.Context) (types.Batch, error) {
+	currentBatchNum, err := k.currentBatchNumber.Peek(ctx)
+	if err != nil {
+		return types.Batch{}, err
+	}
+	batch, err := k.batches.Get(ctx, currentBatchNum-1)
+	if err != nil {
+		return types.Batch{}, err
+	}
```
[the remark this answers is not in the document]
hacheigriega: Agreed. Now the current batch number points to the most recently created batch.

proto/sedachain/batching/v1/batching.proto
```diff
@@ -21,14 +21,17 @@ message Batch {
   // validator_root is the hex-encoded root of the validator merkle
   // tree.
   string validator_root = 4;
+  // batch_id is the Keccack-256 hash of the batch concatenated with
+  // the previous batch ID.
+  bytes batch_id = 5;
```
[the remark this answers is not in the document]
hacheigriega: Not according to this decision https://www.notion.so/sedaprotocol/Batch-Structure-Definition-10da68d575ca806fac07e08b335fca91?pvs=4

app/app.go
```diff
@@ -976,6 +980,33 @@ func NewApp(
 	app.SetEndBlocker(app.EndBlocker)
 	app.SetAnteHandler(anteHandler)
 
+	// Create vote extension for signing batches.
+	pvKeyFile := filepath.Join(homePath, cast.ToString(appOpts.Get("priv_validator_key_file")))
+	loadPath := filepath.Join(filepath.Dir(pvKeyFile), utils.SEDAKeyFileName)
+	signer, err := utils.LoadSEDASigner(loadPath)
+	if err != nil {
+		app.Logger().Error("error loading SEDA signer - ExtendVoteHandler will not run", "path", loadPath)
```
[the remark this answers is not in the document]
hacheigriega: That sounds good to me, but restarting the process to activate the signer seems inevitable, right?

hacheigriega:
> I'm not sold on using the consensus key in the Pubkey registry. I get that it saves a lookup during the vote extension verification but given that the next version of the Cosmos SDK introduces consensus key rotation I think we're setting ourselves up for difficulties.

Good point.. It was a hasty decision. I reverted the change in the pubkey module and added a staking keeper to the vote extension handler.

app/abci/vote_extension.go
hacheigriega: Perhaps it's the error message that should be changed? In our design, a batch will be created at every block, and validators will always be signing a batch created from the previous height, right?

proto/sedachain/batching/v1/batching.proto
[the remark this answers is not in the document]
Thomasvdam: Maybe I’m looking at the wrong section but I think it says in the proposed solution for the Batch ID derivation that we don’t include the previous batch id. 

```
batch_id = keccak256(batch_height, block_height, validator_root, results_root)
```

app/app.go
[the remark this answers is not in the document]
Thomasvdam: It's probably the easiest, but we could (not saying we should :P) do an approach where we always register the handler and have that be responsible for loading the keyfile. When extending a vote it just checks if it has the keys in memory, and if not it attempts to read the file.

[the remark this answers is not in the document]
gluax: We could just have dynamic watching of the file so they don't have to restart? That would mean if that file is edited, changed and saved, it would be reloaded.

scripts/local_multi_setup.sh
```diff
@@ -1,5 +1,9 @@
 #!/bin/bash
 set -e
+set -x
```
gluax, line 3: can do `set -ex`

scripts/local_multi_setup.sh
```diff
@@ -9,132 +13,106 @@ rm -rf $HOME/.sedad/
...
 # copy tendermint node id of validator1 to persistent peers of validator2-4
-sed -i -E "s|persistent_peers = \"\"|persistent_peers = \"$(sedad tendermint show-node-id --home=$HOME/.sedad/validator1)@localhost:26656\"|g" $HOME/.sedad/validator2/config/config.toml
-sed -i -E "s|persistent_peers = \"\"|persistent_peers = \"$(sedad tendermint show-node-id --home=$HOME/.sedad/validator1)@localhost:26656\"|g" $HOME/.sedad/validator3/config/config.toml
-sed -i -E "s|persistent_peers = \"\"|persistent_peers = \"$(sedad tendermint show-node-id --home=$HOME/.sedad/validator1)@localhost:26656\"|g" $HOME/.sedad/validator4/config/config.toml
+sed -i '' -E "s|persistent_peers = \"\"|persistent_peers = \"$($BIN tendermint show-node-id --home=$HOME/.sedad/validator1)@localhost:26656\"|g" $HOME/.sedad/validator2/config/config.toml
 
 # start all four validators
 tmux new-session -s validator1 -d 
-tmux send -t validator1 'sedad start --home=$HOME/.sedad/validator1' ENTER
+tmux send -t validator1 'BIN=./build/sedad' ENTER
```
gluax, line 74: Should be listed as a dev dep for testing IMO. Also would be nice if the script errors if tmux, or any other program needed, isn't installed at the top of the file.

app/abci/vote_extension.go
hacheigriega: Now the batches are primarily indexed by the block height and additionally indexed by the reference key batch number.

proto/sedachain/batching/v1/batching.proto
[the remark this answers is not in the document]
hacheigriega: You're right. Fixed.

app/app.go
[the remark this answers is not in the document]
hacheigriega: interesting ideas.. added an issue here https://github.com/sedaprotocol/seda-chain/issues/375

app/abci/vote_extension.go
```diff
@@ -0,0 +1,74 @@
...
+
+		signature, err := h.signer.Sign(batch.BatchId, utils.Secp256k1)
+		if err != nil {
+			return nil, err
+		}
+
+		h.logger.Debug("submitting batch signature", "signature", signature)
+		return &cometabci.ResponseExtendVote{VoteExtension: signature}, nil
```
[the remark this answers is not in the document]
hacheigriega: I was thinking we would just concatenate the signatures since they should be fixed in length?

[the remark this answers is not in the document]
Thomasvdam: Sounds good! I would suggest to already implement taking the slice from the vote extension in the verification step, even though in practice they're identical right now.

x/batching/keeper/abci.go
```diff
@@ -34,37 +36,64 @@ func (k Keeper) EndBlock(ctx sdk.Context) (err error) {
...
+			return types.Batch{}, err
+		}
+	} else {
+		oldBatchID = oldBatch.BatchId
+	}
+
+	// Compute the batch ID, which is defined as
+	// keccak256(batch_number, block_height, validator_root, results_root).
+	var hashContent []byte
+	hashContent = append(hashContent, oldBatchID...)
```
Thomasvdam, line 80: This is no longer required right?

hacheigriega: phew yes thanks

app/abci/vote_extension.go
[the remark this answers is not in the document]
hacheigriega: good point.. now the verification logic checks `signature[:64]`

app/app.go
```diff
@@ -976,6 +980,33 @@ func NewApp(
...
+		app.Logger().Info("SEDA signer successfully loaded")
+
+		voteExtensionsHandler := appabci.NewVoteExtensionHandler(
+			app.BatchingKeeper,
+			app.PubKeyKeeper,
+			signer,
+			app.Logger(),
+		)
+		app.SetExtendVoteHandler(voteExtensionsHandler.ExtendVoteHandler())
+		app.SetVerifyVoteExtensionHandler(voteExtensionsHandler.VerifyVoteExtensionHandler())
```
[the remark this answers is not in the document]
hacheigriega: technically no because `voteExtensionsHandler` is constructed inside the if..

x/batching/keeper/abci.go
```diff
@@ -74,44 +88,44 @@ func (k Keeper) ConstructBatch(ctx sdk.Context) (types.Batch, error) {
 // ConstructDataResultTree constructs a data result tree based on the
 // data results that have not been batched yet. It returns the tree's
 // entries (unhashed leaf contents) and hex-encoded root.
-func (k Keeper) ConstructDataResultTree(ctx sdk.Context) ([][]byte, string, error) {
+func (k Keeper) ConstructDataResultTree(ctx sdk.Context) ([][]byte, []byte, error) {
```
Thomasvdam, line 91: Not something for this PR, just wanted to get the stupid idea out there. If there are no new DR results does it make sense to just reuse the previous DR results root? As a poor mans batch skip :P

x/batching/keeper/abci.go
```diff
@@ -34,37 +36,49 @@ func (k Keeper) EndBlock(ctx sdk.Context) (err error) {
...
 		return types.Batch{}, err
 	}
-	valEntries, valRootHex, err := k.ConstructValidatorTree(ctx)
+	valEntries, valRoot, err := k.ConstructValidatorTree(ctx)
 	if err != nil {
 		return types.Batch{}, err
 	}
 
+	// Compute the batch ID, which is defined as
+	// keccak256(batch_number, block_height, validator_root, results_root).
```
Thomasvdam, line 64: We're missing the `proving_metadata` or however we decide to name the field, or did we decide against it in the end?

Thomasvdam: Just saw the other PR, I'm ok with adding it there.

hacheigriega: By the way, according to the specs `proving_metadata` is not included when we derive the ID.

Thomasvdam: I think that's because this decision came after that spec. We should check with Mario

gluax (approved):

x/batching/keeper/abci.go
hacheigriega: I'll make sure to address that point

PR #376, opened 2024-10-08
refactor(batching): add separate store for tree entries

hacheigriega:
## Motivation

This PR adds a separate state in the batching module for storing the data result tree and validator tree entries. The entries are returned in base64 format.
Also adds an unused metadata field (type `[]byte`) to the batch type.


x/batching/keeper/data_result.go
Thomasvdam, line 1: Do we already want to add the query `GetDataResultForDataRequestID`? (but with a better name :P)

gluax (approved):

x/batching/keeper/data_result.go
hacheigriega: The problem with `Dr` is that it's short for both `DataRequest` and `DataResult` :/
I think we can just call it `GetDataResult` and take data request ID as an input.

hacheigriega: I will address this in a separate PR. Issue: https://github.com/sedaprotocol/seda-chain/issues/379

Issue #353, opened 2024-09-04
✨ Batch signatures validation and persistence


hacheigriega:
## ✨ Feature

A canonical set of votes that include the batch signatures become available in the next block to the proposer. The last phase of the batching module implementation involves injecting the vote extensions in the proposal in `PrepareProposal` ABCI 2.0 method. Subsequently, the validators will ensure that more than 2/3 of voting power has signed the batch in the `ProcessProposal` method. Finally, batch will be stored in the module store in the `PreBlock` method. Note that once the batch is persisted, the link between the batched data requests and the corresponding batch numbers must be updated, probably through a contract call to the core contract.

PR #377, opened 2024-10-09
feat (batching): batch signature validation and persistence

hacheigriega:
Completion of initial implementation of batching sequence.

## Explanation of Changes

Suppose a batch is created at block height `H`. The newly added `abci.Handlers` process batch signatures in the following sequence:
- `ExtendVoteHandler()` at `H+1`: Handles the ExtendVote ABCI to sign a batch created from the previous block.
- `VerifyVoteExtensionHandler()` at `H+1`: Handles the VerifyVoteExtension ABCI to verify the batch signature included in the pre-commit vote against the public key registered in the pubkey module.
- `PrepareProposalHandler()` at `H+2`: Handles the PrepareProposal ABCI to inject a canonical set of vote extensions in the proposal.
- `ProcessProposalHandler()` at `H+2`: Handles the ProcessProposal ABCI to validate the canonical set of vote extensions injected by the proposer. The validation checks whether the 2/3 voting power has been met and whether the batch signatures are valid.
- `PreBlocker()` at `H+2`: Runs before `BeginBlocker` to extract the batch signatures from the canonical set of vote extensions injected by the proposer and store them.

## Related PRs and Issues

Closes: #353


x/batching/keeper/abci_test.go
```diff
@@ -140,6 +140,7 @@ func addBatchSigningValidators(t *testing.T, f *fixture, num int) ([]sdk.AccAddr
 // generateDataResults returns a given number of randomly-generated
 // data results.
 func generateDataResults(t *testing.T, num int) []types.DataResult {
+	t.Helper()
```
Thomasvdam, line 143: This is cool! Didn't know about that

e2e/e2e_setup_test.go
```diff
@@ -228,6 +228,8 @@ func (s *IntegrationTestSuite) initGenesis(c *chain) {
 	appGenesis.AppState, err = json.MarshalIndent(appGenState, "", "  ")
 	s.Require().NoError(err)
 
+	appGenesis.Consensus.Params.ABCI.VoteExtensionsEnableHeight = 10000
```
Thomasvdam, line 231: Maybe good to add a comment here we set it to disable them for the tests?

app/abci/handlers.go
```diff
@@ -0,0 +1,222 @@
...
+func (h *Handlers) PrepareProposalHandler() sdk.PrepareProposalHandler {
+	return func(ctx sdk.Context, req *abci.RequestPrepareProposal) (*abci.ResponsePrepareProposal, error) {
+		h.logger.Debug("start prepare proposal handler")
+
+		err := baseapp.ValidateVoteExtensions(ctx, h.stakingKeeper, req.Height, ctx.ChainID(), req.LocalLastCommit)
+		if err != nil {
+			return nil, err
+		}
+
+		proposalTxs := req.Txs
```
Thomasvdam, line 100: Should we consider max block size?

app/abci/handlers.go
```diff
@@ -0,0 +1,222 @@
...
+	return func(ctx sdk.Context, req *abci.RequestProcessProposal) (*abci.ResponseProcessProposal, error) {
+		if req.Height > ctx.ConsensusParams().Abci.VoteExtensionsEnableHeight {
+			var extendedVotes abci.ExtendedCommitInfo
+			if err := json.Unmarshal(req.Txs[0], &extendedVotes); err != nil {
+				h.logger.Error("failed to decode injected extended votes tx", "err", err)
+				return &abci.ResponseProcessProposal{Status: abci.ResponseProcessProposal_REJECT}, nil
+			}
+
+			// Validate signatures and perform 2/3 voting power check.
+			err := baseapp.ValidateVoteExtensions(ctx, h.stakingKeeper, req.Height, ctx.ChainID(), extendedVotes)
```
Thomasvdam, line 127: Where is `ValidateVoteExtensions` and how does it do the 2/3rds check? 😬

app/abci/handlers.go
```diff
@@ -0,0 +1,222 @@
...
+			}
+			// Handle an error.
+			if err != nil {
+				h.logger.Error("error in pre-blocker", "err", err)
+			}
+			err = nil
+		}()
+
+		res = new(sdk.ResponsePreBlock)
+		if req.Height > ctx.ConsensusParams().Abci.VoteExtensionsEnableHeight {
```
Thomasvdam, line 169: Maybe an early return makes sense here?

app/abci/handlers.go
```diff
@@ -0,0 +1,222 @@
...
+			h.logger.Debug("begin pre-block logic for batch signature persistence")
+
+			var extendedVotes abci.ExtendedCommitInfo
+			if err := json.Unmarshal(req.Txs[0], &extendedVotes); err != nil {
+				h.logger.Error("failed to decode injected extended votes tx", "err", err)
+				return nil, err
+			}
+
+			// Get batch number of the batch from two blocks ago.
+			batch, err := h.batchingKeeper.GetBatchForHeight(ctx, ctx.BlockHeight()-2)
```
Thomasvdam, line 179: Maybe it's an idea to change the `-1` and `-2` into named constants? We can than add a comment at the constant why we need -1 for the signing and -2 for the prepare/process/preblock

Thomasvdam (commented):
Once we have this in place we should do extensive testing with validators entering and leaving the active set. I'm still a little worried that the delay between batch creation and the voting might cause issues for us.

app/abci/handlers.go
hacheigriega: It's a helper function provided by Cosmos https://github.com/cosmos/cosmos-sdk/blob/8bfcf554275c1efbb42666cc8510d2da139b67fa/baseapp/abci_utils.go#L44

I got a question from re-reading this code. It seems the voting power is not included in the sign message? Not sure how we can trust the power included in the vote if this is the case.

app/abci/handlers.go
hacheigriega: Where?

Thomasvdam: Instead of having all the relevant code in the `if` clause we can do 

```golang
if req.Height <= ctx.ConsensusParams().Abci.VoteExtensionsEnableHeight {
  return res, nil
}
```

app/abci/handlers.go
Thomasvdam: Aaah, I thought the comment referred to our 2/3rds check for the batch, but if this code verifies that the received votes all have valid extensions and total more than 2/3rds it should also work for our batches.

gluax (approved):
Approving but Thomas has some nice readability suggestions we should do IMO :)

x/tally/types/filters.go
```diff
@@ -3,9 +3,9 @@ package types
 import (
 	"bytes"
 	"encoding/binary"
-	"slices"
 
 	"golang.org/x/exp/constraints"
+	"slices"
```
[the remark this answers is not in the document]
hacheigriega: I know.. some weird linter issue. I will look into it

app/abci/handlers.go
hacheigriega: Now the handlers wrap the default proposal handlers and run them after the vote injection logic.

app/abci/handlers.go
hacheigriega: updated

app/abci/handlers.go
```diff
@@ -0,0 +1,270 @@
...
+}
+
+// VerifyVoteExtensionHandler handles the VerifyVoteExtension ABCI to
+// verify the batch signature included in the pre-commit vote against
+// the public key registered in the pubkey module.
+func (h *Handlers) VerifyVoteExtensionHandler() sdk.VerifyVoteExtensionHandler {
+	return func(ctx sdk.Context, req *abci.RequestVerifyVoteExtension) (*abci.ResponseVerifyVoteExtension, error) {
+		h.logger.Debug("start verify vote extension handler", "request", req)
+
+		if len(req.VoteExtension) != 64 {
```
Thomasvdam, line 89: I think it makes sense to put a cap on the vote extension size, but if we make it the exact size of our signatures we're not forward compatible. Maybe it makes sense to reserve space for ~5 signatures? So we don't immediately run into problem when introducing new schemes.

app/abci/handlers.go
```diff
@@ -0,0 +1,270 @@
...
+		var injection []byte
+		if req.Height > ctx.ConsensusParams().Abci.VoteExtensionsEnableHeight {
+			err := baseapp.ValidateVoteExtensions(ctx, h.stakingKeeper, req.Height, ctx.ChainID(), req.LocalLastCommit)
+			if err != nil {
+				return nil, err
+			}
+
+			injection, err = json.Marshal(req.LocalLastCommit)
+			if err != nil {
+				h.logger.Error("failed to marshal extended votes", "err", err)
```
Thomasvdam, line 123: Should we not abort if this fails?

app/abci/handlers.go
```diff
@@ -0,0 +1,270 @@
...
+			}
+
+			injectionSize := int64(len(injection))
+			if injectionSize <= req.MaxTxBytes {
+				req.MaxTxBytes -= injectionSize
+			} else {
```
Thomasvdam, line 129: Personal preference but we can also use an early return here to drop the `else` clause :)

app/abci/handlers.go
```diff
@@ -0,0 +1,270 @@
...
+					"injection_size", injectionSize,
+					"MaxTxBytes", req.MaxTxBytes,
+				)
+				return &abci.ResponsePrepareProposal{Txs: make([][]byte, 0)}, ErrVoteExtensionInjectionTooBig
+			}
+		}
+
+		defaultRes, err := h.defaultPrepareProposal(ctx, req)
+		if err != nil {
+			h.logger.Error("failed to run default prepare proposal handler", "err", err)
```
Thomasvdam, line 141: Same here, is this not something that should bubble up?

app/abci/handlers.go
```diff
@@ -0,0 +1,270 @@
...
+			Txs: proposalTxs,
+		}, nil
+	}
+}
+
+// ProcessProposalHandler handles the ProcessProposal ABCI to validate
+// the canonical set of vote extensions injected by the proposer.
+func (h *Handlers) ProcessProposalHandler() sdk.ProcessProposalHandler {
+	return func(ctx sdk.Context, req *abci.RequestProcessProposal) (*abci.ResponseProcessProposal, error) {
+		if req.Height > ctx.ConsensusParams().Abci.VoteExtensionsEnableHeight {
```
Thomasvdam, line 158: Broken record time: early return? 😅

app/abci/handlers.go
```diff
@@ -0,0 +1,270 @@
...
+				return nil, err
+			}
+
+			// Verify every batch signature.
+			batch, err := h.batchingKeeper.GetBatchForHeight(ctx, ctx.BlockHeight()+proposalOffset)
+			if err != nil {
+				return nil, err
+			}
+			for _, vote := range extendedVotes.Votes {
+				if len(vote.VoteExtension) != 64 {
```
Thomasvdam, line 177: We should probably assign the 64s to a named constant. Same for the max vote length.

app/abci/handlers.go
```diff
@@ -0,0 +1,270 @@
...
+					h.logger.Error("invalid length vote extension in proposal", "vote", vote)
+					return nil, ErrInvalidVoteExtensionLength
+				}
+
+				ok, err := h.verifyBatchSignature(ctx, batch.BatchId, vote.VoteExtension[:64], vote.Validator.Address, utils.Secp256k1)
+				if !ok || err != nil {
+					return nil, err
```
Thomasvdam, line 184: For the `!ok` case won't the SDK be confused it it receives a `nil` value for both the `ResponseProcessProposal` and the error?

app/abci/handlers.go
hacheigriega: really good point. I added `const maxVoteExtensionLength = 64*5`. Also moved this check into `verifyBatchSignatures()`.

app/abci/handlers.go
hacheigriega: Yeah good point. This was a bad mistake. Added a new error `ErrInvalidBatchSignature` to address this.

On a different but related note, I am still confused by how Cosmos is handling errors from these ABCI methods. I find this issue particularly confusing https://github.com/cosmos/cosmos-sdk/pull/17930. Gotta look into it more and test things out.

app/abci/handlers.go
hacheigriega: yeah bad mistake. fixed

hacheigriega:
The interchaintests are fixed in #381.

Issue #371, opened 2024-10-03
✨ Publish Data Request Result events


Thomasvdam:
## ✨ Feature

With #355 we're moving Data Request Results from the core contract to a module. We will have to update the indexing logic to account for this change.

## Motivation

Keep the data request indexing functionality intact.

## Implementation

 - [ ] Create overview of what data is available in terms of the store and emitted events.
 - [ ] Create new module in indexing plugin that publishes result events.

PR #382, opened 2024-10-15
Publish data result events

Thomasvdam:
## Motivation

Support indexing results.

## Explanation of Changes

Reimplement the struct to avoid having to deal with the `omitempty` tags on the receiving end.

## Testing

Add code in `x/tally/keeper/abci.go` that inserts a dummy result every block.

Run the chain and plugin locally with a queue emulator, messages get published with the following JSON:

```json
{
	"type": "data-result",
	"data": {
                "result_id": "74",
		"dr_id": "74",
		"version": "1",
		"block_height": 74,
		"exit_code": 0,
		"gas_used": 0,
		"result": "eyJyZXN1bHQiOiAiMSJ9",
		"payback_address": "1",
		"seda_payload": "1",
		"consensus": true
	}
}
```

## Related PRs and Issues

Closes: #371

### DIFF
--- a/plugins/indexing/batching/module.go
+++ b/plugins/indexing/batching/module.go
@@ -1,0 +1,59 @@
+package batching
+
+import (
+	"bytes"
+
+	// "cosmossdk.io/collections"
+	storetypes "cosmossdk.io/store/types"
+
+	"github.com/cosmos/cosmos-sdk/codec"
+
+	"github.com/sedaprotocol/seda-chain/plugins/indexing/log"
+	"github.com/sedaprotocol/seda-chain/plugins/indexing/types"
+	batchingtypes "github.com/sedaprotocol/seda-chain/x/batching/types"
+)
+
+const StoreKey = batchingtypes.StoreKey
+
+func ExtractUpdate(ctx *types.BlockContext, cdc codec.Codec, logger *log.Logger, change *storetypes.StoreKVPair) (*types.Message, error) {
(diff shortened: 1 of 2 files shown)


plugins/indexing/batching/module.go
```diff
@@ -0,0 +1,59 @@
...
+			ID             string `json:"result_id"`
+			DrID           string `json:"dr_id"`
+			Version        string `json:"version"`
+			BlockHeight    uint64 `json:"block_height"`
+			ExitCode       uint32 `json:"exit_code"`
+			GasUsed        uint64 `json:"gas_used"`
+			Result         []byte `json:"result"`
+			PaybackAddress string `json:"payback_address"`
+			SedaPayload    string `json:"seda_payload"`
+			Consensus      bool   `json:"consensus"`
```
gluax, line 40: Would it make more sense to have this as a proper structure if we need to use it outside this instance? Or is it only used in this instance?

gluax (approved):

hacheigriega (approved):

plugins/indexing/batching/module.go
hacheigriega: Oh true. Why don't we just use `DataResult` type defined in `x/batching/types/batching.pb.go`?

Thomasvdam: Those have the omitempty attribute and I’d prefer not to have to deal with that at the receiving end 😅

hacheigriega: Ah yes, you put it in the prompt 🙃

Issue #378, opened 2024-10-10
♻️ Rename data request wasm to oracle program


hacheigriega:
## ♻️ Refactor

The wasm storage makes a distinction between `DataRequestWasm` and `ExecutorWasm`. We intend to rename the former to `OracleProgram`.

PR #393, opened 2024-10-22
refactor(x/wasm-storage): rename data request wasm to oracle program

hacheigriega:
## Explanation of Changes

"Data request wasm" has been renamed to "oracle program." This change is both API breaking and state machine breaking.

```
sedad query wasm-storage data-request-wasm <hash>
sedad query wasm-storage list-data-request-wasms
sedad tx wasm-storage store-data-request-wasm <wasm file>

->

sedad query wasm-storage oracle-program <hash>
sedad query wasm-storage list-oracle-programs
sedad tx wasm-storage store-oracle-program <wasm file>
```

Swagger API docs have been updated to reflect the changes in the REST API endpoints.

## Related PRs and Issues

Closes: #378


client/docs/swagger-ui/swagger.yaml
```diff
@@ -41,10 +41,10 @@ paths:
                       format: byte
       tags:
         - Query
-  /seda-chain/wasm-storage/data_request_wasm/{hash}:
+  /seda-chain/wasm-storage/executor_wasm/{hash}:
```
gluax, line 44: `oracle_program` no? They aren't specific to execution since they can also be used for tallying.

Unless these are the executor WASMs for the overlay to use? Maybe we could update the description to say it's for overlay use?

client/docs/swagger-ui/swagger.yaml
```diff
@@ -31934,14 +31928,20 @@ definitions:
...
+    description: OracleProgram is a wasm used for data request.
   sedachain.wasm_storage.v1.QueryCoreContractRegistryResponse:
     type: object
     properties:
       address:
         type: string
     description: The response message for QueryCoreContractRegistry RPC.
-  sedachain.wasm_storage.v1.QueryDataRequestWasmResponse:
+  sedachain.wasm_storage.v1.QueryExecutorWasmResponse:
     type: object
```
gluax, line 31945: Maybe QueryDataRequestResponse would be better? Or is this the overlay Executor WASM again?

e2e/e2e_gov_test.go
```diff
@@ -32,7 +32,7 @@ func (s *IntegrationTestSuite) testWasmStorageStoreExecutorWasm() {
 
 	bytecode, err := os.ReadFile(filepath.Join(localWasmDirPath, executorWasm))
 	if err != nil {
-		panic("failed to read data request Wasm file")
+		panic("failed to read a wasm file")
```
gluax, line 35: Unsure if we should say `oracle program` here as well. As if a user gets an error they might got `wtf is a wasm`.

e2e/e2e_register_proxy_contract_test.go
```diff
@@ -142,7 +142,7 @@ func (s *IntegrationTestSuite) execWasmStore(
 		command = append(command, fmt.Sprintf("--%s=%v", flag, value))
 	}
 
-	s.T().Logf("storing data request Wasm %s on chain %s", wasmFilePath, c.id)
+	s.T().Logf("storing a wasm file %s on chain %s", wasmFilePath, c.id)
```
gluax, line 145: Same as above.

x/wasm-storage/keeper/keeper.go
```diff
@@ -199,7 +199,7 @@ func (k Keeper) GetAllExecutorWasms(ctx sdk.Context) []types.ExecutorWasm {
 	return wasms
 }
 
-// GetExpiredWasmKeys retrieves the keys of the data request wasms that
+// GetExpiredWasmKeys retrieves the keys of the oracle programs that
 // will expire at the given block height.
 func (k Keeper) GetExpiredWasmKeys(ctx sdk.Context, expirationHeight int64) ([][]byte, error) {
```
gluax, line 204: Is this internal only? Otherwise I'd say `GetExpiredOracleProgamKeys`

gluax (commented):
A few questions/comments. Otherwise looks good to me, but I think waiting for a second set of eyes/opinions would be good.

x/tally/keeper/abci.go
```diff
@@ -208,7 +208,7 @@ func (k Keeper) FilterAndTally(ctx sdk.Context, req types.Request) (TallyResult,
 		return result, errorsmod.Wrap(err, "error while applying filter")
 	}
 
-	tallyWasm, err := k.wasmStorageKeeper.GetDataRequestWasm(ctx, req.TallyBinaryID)
+	tallyWasm, err := k.wasmStorageKeeper.GetOracleProgram(ctx, req.TallyBinaryID)
```
Thomasvdam, line 211: Should this not be `req.TallyProgramID`?

x/wasm-storage/keeper/keeper.go
```diff
@@ -148,11 +148,11 @@ func (k Keeper) IterateExecutorWasms(ctx sdk.Context, callback func(wasm types.E
 	return nil
 }
 
-// ListDataRequestWasms returns hashes and expiration block heights
-// of all data request wasms in the store.
-func (k Keeper) ListDataRequestWasms(ctx sdk.Context) []string {
+// ListOraclePrograms returns hashes and expiration block heights
+// of all oracle programs in the store.
+func (k Keeper) ListOraclePrograms(ctx sdk.Context) []string {
```
Thomasvdam, line 153: Should we add pagination here? Given that we have to load the bytecode for all Oracle Programs I imagine the load might increase quite rapidly.

Thomasvdam (approved):

client/docs/swagger-ui/swagger.yaml
hacheigriega: I think this is just git diff detecting wrong lines

client/docs/swagger-ui/swagger.yaml
hacheigriega: In reality `QueryDataRequestWasmResponse` was renamed to `QueryOraclePrograms`.

x/tally/keeper/abci.go
Thomasvdam: There are only oracle programs, it’s just that in the DR you can specify one for the execution phase `exec_program_id` and the tally phase `tally_program_id`

x/wasm-storage/keeper/keeper.go
hacheigriega: good idea.. done

Issue #392, opened 2024-10-21
♻️ Skip batches with no updates


hacheigriega:
## ♻️ Refactor

When a block does not contain any new data results nor any changes in the active validator set, the resulting batch's tree roots are identical to the previous batch's. In such a scenario, we can skip batching for the block altogether.

PR #394, opened 2024-10-22
feat(x/batching): skip batches with no updates

hacheigriega:
## Explanation of Changes

- We skip batching if the validator root matches the previous batch's AND there are no new data results to ready be batched. Alternatively, we could check for validator updates through the staking keeper instead of generating a validator tree every block, but this will fail to detect a change in a validator's public key. 
- A batch now has a new field `CurrentDataResultRoot`, which is the root of the merkle tree constructed from the current block's batch-ready data results. The field `DataResultRoot`, as before, is the "super root" of `CurrentDataResultRoot` and the previous batch's `DataResultRoot`.
- We now store the tree entries with their domain separators removed.
- Validator tree entries now contain Ethereum-style addresses (last 20 bytes of Keccak 256 hash of a given public key). 
- The voting power field has been removed from `BatchSignature`. 

## Related PRs and Issues

Closes: #392


x/batching/keeper/abci_test.go
```diff
@@ -143,15 +140,17 @@ func addBatchSigningValidators(t *testing.T, f *fixture, num int) ([]sdk.AccAddr
 	return addrs, pubKeys, powers
 }
 
-// decompressPubKey decompresses a 33-byte long compressed public key
-// into a 64-byte long uncompressed format.
-func decompressPubKey(t *testing.T, pubKey []byte) []byte {
+// compressedPubKeyToAddress converts a compressed public key to an
+// Ethereum-style address.
+func compressedPubKeyToAddress(t *testing.T, pubKey []byte) []byte {
```
Thomasvdam, line 145: Should this not be the same implementation as what we're using in the utils?

x/batching/keeper/endblock.go
```diff
@@ -49,21 +52,44 @@ func (k Keeper) EndBlock(ctx sdk.Context) (err error) {
...
 	}
-	newBatchNum := oldBatchNum + 1
 
+	// Compute current data result tree root and the "super root"
+	// of current and previous data result trees' roots.
 	dataEntries, dataRoot, err := k.ConstructDataResultTree(ctx, newBatchNum)
 	if err != nil {
 		return types.Batch{}, nil, nil, err
 	}
+	latestDataRootHex, err := hex.DecodeString(latestDataRoot)
```
Thomasvdam, line 75: `latestDataRootHex` is misleading right? It's just bytes at this point.

x/batching/keeper/endblock.go
```diff
@@ -49,21 +52,44 @@ func (k Keeper) EndBlock(ctx sdk.Context) (err error) {
...
 	if err != nil {
-		return types.Batch{}, nil, nil, err
+		if !errors.Is(err, types.ErrBatchingHasNotStarted) {
+			return types.Batch{}, nil, nil, err
+		}
+		newBatchNum = collections.DefaultSequenceStart + 1
+	} else {
+		newBatchNum = latestBatch.BatchNumber + 1
+		latestDataRoot = latestBatch.DataResultRoot
+		latestValRoot = latestBatch.ValidatorRoot
```
Thomasvdam, line 66: Maybe it makes sense to sufffix these with `Hex`?

x/batching/keeper/endblock.go
```diff
@@ -49,21 +52,44 @@ func (k Keeper) EndBlock(ctx sdk.Context) (err error) {
...
 	if err != nil {
 		return types.Batch{}, nil, nil, err
 	}
+	valRootHex := hex.EncodeToString(valRoot)
+
+	// Skip batching if there is no update in data result root nor
+	// validator root.
+	if len(dataEntries) == 0 && valRootHex == latestValRoot {
+		return types.Batch{}, nil, nil, types.ErrNoBatchingUpdate
+	}
```
Thomasvdam, line 92: Not sure if it's worth it but we could move the DR root construction after this check.

x/batching/keeper/endblock.go
```diff
@@ -79,59 +105,56 @@ func (k Keeper) ConstructBatch(ctx sdk.Context) (types.Batch, [][]byte, [][]byte
...
 func (k Keeper) ConstructDataResultTree(ctx sdk.Context, newBatchNum uint64) ([][]byte, []byte, error) {
 	dataResults, err := k.GetDataResults(ctx, false)
 	if err != nil {
 		return nil, nil, err
 	}
 
 	entries := make([][]byte, len(dataResults))
+	treeEntries := make([][]byte, len(dataResults))
 	for i, res := range dataResults {
 		resHash, err := hex.DecodeString(res.Id)
```
Thomasvdam, line 130: Maybe it's helpful to name this `resID`? I get that it's the hash of the result, but it might help to avoid confusion whether the result ID is hashed again or not.

x/batching/keeper/endblock.go
```diff
@@ -142,7 +165,7 @@ func (k Keeper) ConstructValidatorTree(ctx sdk.Context) ([][]byte, []byte, error
 			}
 			panic(err)
 		}
-		pkBytes, err := decompressPubKey(secp256k1PubKey.Bytes())
+		addr, err := utils.PubKeyToAddress(secp256k1PubKey.Bytes())
```
Thomasvdam, line 168: I'm probably being overly verbose but I'd consider naming this something like `ethAddr`, since in this Cosmos context it's reasonable to expect it to be a bech32 address.

x/batching/keeper/endblock.go
```diff
@@ -153,31 +176,20 @@ func (k Keeper) ConstructValidatorTree(ctx sdk.Context) ([][]byte, []byte, error
 
 		// TODO Validator set trimming
 
-		// An entry is (domain_separator || pubkey || voting_power_percentage).
-		entry := make([]byte, len(separator)+len(pkBytes)+4)
-		copy(entry[:len(separator)], separator)
-		copy(entry[len(separator):len(separator)+len(pkBytes)], pkBytes)
+		// An entry is (domain_separator | address | voting_power_percentage).
```
Thomasvdam, line 179: A tree entry*

app/abci/handlers.go
```diff
@@ -230,29 +265,31 @@ func (h *Handlers) PreBlocker() sdk.PreBlocker {
 			return res, nil
 		}
 
-		h.logger.Debug("begin pre-block logic for batch signature persistence")
+		batch, err := h.batchingKeeper.GetBatchForHeight(ctx, ctx.BlockHeight()+BlockOffsetCollectPhase)
+		if err != nil {
+			if errors.Is(err, collections.ErrNotFound) {
+				h.logger.Debug("no batch to collect signatures for", "height", ctx.BlockHeight())
+				return nil, nil
```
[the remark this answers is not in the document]
hacheigriega: Probably doesn't matter but still a good catch. fixed.

x/batching/keeper/endblock.go
hacheigriega: oops good catch. fixed.

x/batching/keeper/endblock.go
hacheigriega: that's what I did before seeing this comment !

x/batching/keeper/endblock.go
hacheigriega: I don't think that's worth it because if there are no data results to be batched, the method `ConstructDataResultTree` would be very light anyways.

x/batching/keeper/endblock.go
hacheigriega: good idea, I also changed the method name to `PubKeyToEthAddress`

x/batching/keeper/endblock.go
hacheigriega: Oh you mean `resHash` -> `resID`, right? Agreed fixed

app/utils/seda_keys.go
```diff
@@ -170,3 +173,19 @@ func (s *sedaKeys) Sign(input []byte, index SEDAKeyIndex) ([]byte, error) {
 	}
 	return signature, nil
 }
+
+// PubKeyToAddress converts a public key in the 33-byte compressed
+// format into the Ethereum address format, which is defined as the
+// rightmost 160 bits of a Keccak hash of an ECDSA public key.
+func PubKeyToAddress(pubkey []byte) ([]byte, error) {
```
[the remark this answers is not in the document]
Thomasvdam: I just thought it was easier since we're already importing the library anyway, didn't realise those functions came with additional CGO baggage :)

Thomasvdam (approved):

Issue #395, opened 2024-10-23
🐛 There are Two Instances of the cosmwasm VM running


gluax:
## 🐛 Bug Report

There are two instances of the `wasmVM` for `cosmwasm` running.

## Steps to Reproduce

It's pretty un-noticeable in the current version of the chain. 
However, upgrading  `wasmd` and `wasmVM` to the versions that use the `2.0` version makes it clear when running the chain and trying to upload a contract.

The newer version prevents 2 VM instances from running in the same directory(from migration guide, as well as changelog), meaning we are having two instances running, just the older version doesn't prevent us from doing this.

#### Code snippet to reproduce

1. Start the chain(after upgrading `wasmd` and `wasmvm`).
2. Try uploading a contract.

#### Stack trace & error message

```
panic: Could not lock exclusive.lock. Is a different VM running in the same directory already?

      goroutine 1 [running]:
      github.com/CosmWasm/wasmd/x/wasm/keeper.NewKeeper({_, _}, {_, _}, {_, _}, {_, _}, {0x43071d0, 0xc0001b7bf8}, ...)
        github.com/CosmWasm/wasmd@v0.53.0/x/wasm/keeper/keeper_cgo.go:75 +0xed2
      github.com/sedaprotocol/seda-chain/app.NewApp({0x43070b0, 0x62771a0}, {0x431e280, 0xc0012d9320}, {0x0, 0x0}, 0x1, 0xc0013c7a10, {0xc0013a4558, 0x12}, ...)
        github.com/sedaprotocol/seda-chain/app/app.go:587 +0x38f4
      github.com/sedaprotocol/seda-chain/cmd/sedad/cmd.NewRootCmd()
        github.com/sedaprotocol/seda-chain/cmd/sedad/cmd/root.go:71 +0x60f
      main.main()
        github.com/sedaprotocol/seda-chain/cmd/sedad/main.go:15 +0x17
```

## Expected Behavior

Uploading the contract should not panic cause there should only be one instance of the VM running.

This could be why #385 is not working as well. Maybe `wasmer` no longer likes/allows 2 instances of a running VM to touch the same place in memory(on specific architectures??). Not sure TBH.

gluax:
I did a bandaid test if doing this would fix #385 without upgrading to `cosmwasm 2.0`, sadly it did not fix that issue :c

hacheigriega:
[...]

PR #398, opened 2024-10-24
feat: update cosmwasm

gluax:
## Motivation

To update `cosmwasm` dependencies.

## Explanation of Changes

Changes the `wasmCapabilities` function to return the array.
Also, band-aids #395, by just using a temporary directory for the temporary app.

We were also getting deprecation warnings for some actions that they will be dead at the end of the month so I updated all the actions I could.

## Testing

Tests pass.

## Related PRs and Issues

N/A


hacheigriega:
Looks great! How come contracts don't have to be upgraded, though? What's the point of migration instructions like this then? https://github.com/CosmWasm/cosmwasm/blob/v2.0.0/MIGRATING.md

gluax:
> Looks great! How come contracts don't have to be upgraded, though? What's the point of migration instructions like this then? https://github.com/CosmWasm/cosmwasm/blob/v2.0.0/MIGRATING.md

I believe it's because you can specify `wasmCapabilities` to still have support for 1.x versions. Not sure if it's entirely backwards compatible, or if it does some other black magic.

However, the migration guide in the cosmwasm rust SDK made no breaking changes. Just interface changes. Like no longer being able to construct a type directly but using methods to create an instance.

.github/workflows/build.yml
```diff
@@ -3,14 +3,12 @@ name: Build
 
 on:
   pull_request:
-    push:
-      branches:
-        - main
+  push:
+    branches:
+      - main
```
Thomasvdam, line 8: This changes the functionality I think: it'll now also start running on pushes to main. Is that desired?

Thomasvdam (commented):
Is it a problem that the simulation fails with the VM lockfile error?

```sh
panic: Could not lock exclusive.lock. Is a different VM running in the same directory already? [recovered]
	panic: Could not lock exclusive.lock. Is a different VM running in the same directory already?
```

gluax:
> Is it a problem that the simulation fails with the VM lockfile error?
> 
> ```shell
> panic: Could not lock exclusive.lock. Is a different VM running in the same directory already? [recovered]
> 	panic: Could not lock exclusive.lock. Is a different VM running in the same directory already?
> ```

Ah, yea, this is fixable. The recommended way from the cosmwasm team is to use a tempdir for every instance of the VM in testing. I'll take a look at that today.

.github/workflows/build.yml
gluax: Hm you're right it would now trigger on pushes to `main`, but before it wasn't doing anything. `push` is not a valid key under `pull_request` my github actions linter says. The [documentation ](https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#pull_request) also seems to back this up.

So I think the `push` section should just be removed.

Thomasvdam (approved):

PR #413, opened 2024-11-05
refactor(abci): Batch signing and signature verification refactor

### LINKED ISSUE #375: ♻️ SEDA signer registration without restart
## ♻️ Refactor

Currently the application at startup reads the SEDA key file and registers the `ExtendVote` ABCI method only if the read is successful. If the SEDA key file is generated after the startup, the application has to be restarted. To avoid the need for restarting the app, we could

- Check for the SEDA key file every time the `ExtendVote` method runs if the signer is `nil`.
- Or "dynamically watch" the file to support key rotation without a restart as well

The issue also calls for clarifying the SEDA key file configurations. Currently the application assumes that the key file is placed in the same directory as the consensus key file.

### LINKED ISSUE #412: ♻️ Resolve inconsistency between chain and prover contract in batch signing logic
## ♻️ Refactor

It came to our attention that the mismatch between batch validation logic on the prover side and the chain side could lead to violation of our assumption that as long as the solver has the complete sequence of batches from the chain, it can update the prover contract to the latest batch. Specifically, whereas the prover contract checks the signatures of batch `n` against the public keys from batch `n-1`, the chain currently assumes that its batch signatories will sign batch `n` with the keys registered in the pubkey module at the time of signing. A problem could emerge when a validator rotates the batching key. An example scenario:

- Validator Valerie signs batch `n` with key A and then rotates her batching key to B.
- Valerie signs batch `n+1` with B.
- Upon receiving batch `n+1`, the prover contract checks Valerie's signature of batch `n+1` against her public key included in batch `n`'s validator tree, which in this case is A. Therefore, Valerie's signature of `n+1` is invalid on the prover side.

Therefore, while we should keep the current batch construction logic, we need some changes in batch signing and batch signature verification logic. We need to require the validators to sign one more batch with their old keys after they rotate their keys. Incidentally, the change in the batch signing logic in `ExtendVote` also provides a clean solution to the issue #375. The propose logic is as follows:

- Look up `x/pubkey` to obtain the currently registered public key
  - If no public key is found, do not sign and return.
  - Check if the retrieved public key is the same as the public key stored in the signer object
    - If the same, sign and return
    - If different, sign with the key stored in the signer object, reload the signer object, and return

In addition, the batch signature validation logic should also be updated so that they are checked against the public key in the previous batch's validator tree, instead of the public key stored in the pubkey module at the current height. If a previous batch does not exist, or if the given validator was not in the previous batch's validator tree, then the public key module should be looked up.

Also related to #391 as this affects the validator onboarding flow.

hacheigriega:
## Motivation

This PR updates the batch signing and verification logic so that it aligns with the prover's logic. We now use the validator tree from the previous batch as the basis for trusting a new batch. If, however, there was no previous batch, the new batch is assumed to be trustworthy. 

## Explanation of Changes

The revised batch sign / verification logic:
Signing in `ExtendVote`
```
Check if there is a batch to sign
	- If false, Return

Sign with currently loaded keys

Check if uploaded public keys match currently loaded keys
	- If true, Return
	- if false, Reload the keys
```

Verifying in `VerifyVoteExtension`
```
Check if there is a batch to sign
	- If false, Return

Check if very first batch
	- If true
		- Verify signatures against public keys in x/pubkey
	- If false
		- Verify signatures against addresses in previous batch
			- If fail // This means it's a new active validator
				- Check that vote extension is empty
```

 This PR also includes a refactor of `treeEntries` in the batching module store. 
```
Before:
batch_number -> TreeEntries

After:
batch_number | ValidatorAddress -> TreeEntry
```

## Related PRs and Issues

Closes #412 
Closes #375


app/app.go
```diff
@@ -1003,10 +1003,9 @@ func NewApp(
 
 	// Register SEDA signer and ExtendVote handler.
 	pvKeyFile := filepath.Join(homePath, cast.ToString(appOpts.Get("priv_validator_key_file")))
-	loadPath := filepath.Join(filepath.Dir(pvKeyFile), utils.SEDAKeyFileName)
-	signer, err := utils.LoadSEDASigner(loadPath)
+	signer, err := utils.LoadSEDASigner(pvKeyFile)
```
Thomasvdam, line 1006: Stupid idea time!

Given that we're expecting validators to create the keyfile with a transaction from the validator operator account, can we not derive the operator address at that point in time and store it in the keyfile? That way we're less dependent on the priv key file and we don't have to do a lookup by consensus address. Also reduces the risk of issues when we upgrade the SDK to a version that has consensus key rotation.

app/utils/seda_signer.go
```diff
@@ -0,0 +1,136 @@
...
+	var keys []indexedPrivKey
+	err = json.Unmarshal(keysJSONBytes, &keys)
+	if err != nil {
+		return fmt.Errorf("failed to unmarshal SEDA keys from %v: %v", s.keyPath, err)
+	}
+	keysMap := make(map[SEDAKeyIndex]indexedPrivKey)
+	for _, key := range keys {
+		keysMap[key.Index] = key
+	}
+	s.keys = keysMap
```
Thomasvdam, line 134: We should also rebuild the `pubKeys` slice right? Since that's used for the mismatch check. Maybe good to extract the load logic into a helper function that both `reload` and `LoadSEDASigner` can use.

Thomasvdam (commented):
I still need to check `app/abci/handlers.go` but I need to go to bed so here's the first couple of comments :')

app/app.go
hacheigriega: if this is your stupid idea, how good are your smart ideas? 😱

app/utils/seda_signer.go
hacheigriega: Good catch.. fixed in https://github.com/sedaprotocol/seda-chain/pull/413/commits/19bc687778fe21ed0be7dc09e3a6c31abf075c6f

app/abci/handlers.go
```diff
@@ -327,15 +343,46 @@ func (h *Handlers) verifyBatchSignatures(ctx sdk.Context, batchID, voteExtension
...
+				}
+				expAddr, err = utils.PubKeyToEthAddress(pubKey)
+				if err != nil {
+					return err
+				}
+			} else {
+				return err
+			}
+		} else {
+			expAddr = entry[:20]
```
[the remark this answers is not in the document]
hacheigriega: good point. will address in that PR!

app/app.go
hacheigriega: addressed in https://github.com/sedaprotocol/seda-chain/pull/413/commits/897dd27f33dc19e38fc9b5e96f5ed828623289ce

app/abci/handlers.go
```diff
@@ -88,10 +88,25 @@ func (h *Handlers) ExtendVoteHandler() sdk.ExtendVoteHandler {
...
+		if err != nil {
+			return nil, err
+		}
+		valKeys, err := h.pubKeyKeeper.GetValidatorKeys(ctx, val.OperatorAddress)
+		if err != nil {
+			return nil, err
+		}
+
+		// Sign and reload the signer if the public key has changed.
 		signature, err := h.signer.Sign(batch.BatchId, utils.SEDAKeyIndexSecp256k1)
```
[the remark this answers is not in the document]
hacheigriega: So `ExtendVote` is run only when the validator is active and is voting for the block. I don't see a case where a validator would be voting but would not be expected to sign the batch corresponding to the block?

app/abci/handlers.go
```diff
@@ -327,15 +343,46 @@ func (h *Handlers) verifyBatchSignatures(ctx sdk.Context, batchID, voteExtension
...
+		if err != nil {
+			if errors.Is(err, collections.ErrNotFound) {
+				pubKey, err := h.pubKeyKeeper.GetValidatorKeyAtIndex(ctx, valOper, utils.SEDAKeyIndexSecp256k1)
+				if err != nil {
+					return err
+				}
+				expAddr, err = utils.PubKeyToEthAddress(pubKey)
+				if err != nil {
+					return err
+				}
```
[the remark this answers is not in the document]
hacheigriega: Yeah really important point.  I came across a bug where a new validator is missing from the validator tree if its public keys have not been registered in the pubkey module (see https://github.com/sedaprotocol/seda-chain/blob/8c097d0e7404130571a93ee00f88f409bdf51f91/x/batching/keeper/endblock.go#L174) 
I will address this along with the last comment tomorrow.

app/utils/seda_signer.go
```diff
@@ -0,0 +1,133 @@
...
+	keyPath  string
+	isLoaded bool
+}
+
+// LoadSEDASigner loads the SEDA keys from a given file path and
+// returns a SEDASigner interface.
+func LoadSEDASigner(keyFilePath string) (SEDASigner, error) {
+	keys, err := loadSEDAKeys(keyFilePath)
+	if err != nil {
+		keys.keyPath = keyFilePath
```
Thomasvdam, line 37: Why do we do this here again? Should be the same as what is set on line 66 right?

x/pubkey/client/cli/tx.go
```diff
@@ -50,11 +51,11 @@ func AddKey(ac address.Codec) *cobra.Command {
 			}
 			serverCfg := server.GetServerContextFromCmd(cmd).Config
 
-			fromAddr := clientCtx.GetFromAddress()
-			if fromAddr.Empty() {
+			valAddr := sdk.ValAddress(clientCtx.GetFromAddress())
```
Thomasvdam, line 54: I'm confused how we can change the address format here in the CLI, but we don't need changes in the msg_server of the Pubkey module?

app/abci/handlers.go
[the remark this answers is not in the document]
Thomasvdam: That makes sense though right? If a validator doesn't have keys registered they should never be able to participate in the batch signing process, so it doesn't make sense to have them as part of the tree.

I think this is part of the onboarding/bootstrapping problem, if we think it poses an issue we can make a note and address it in a future PR that deals with the bootstrapping phase.

app/utils/seda_signer.go
hacheigriega: Not if `loadSEDAKeys()` results in an error though.

app/abci/handlers.go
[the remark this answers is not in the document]
hacheigriega: added a check in https://github.com/sedaprotocol/seda-chain/pull/413/commits/2c163bccc3697be182f8ac089ee804d8e00d2a1b

app/abci/handlers.go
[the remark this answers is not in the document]
hacheigriega: Now we check if the signature is empty in this case

x/pubkey/client/cli/tx.go
hacheigriega: So this is just wrapping the underlying bytes. I suppose we could use the bytes directly for the message, but that would reduce readability.

app/abci/handlers.go
```diff
@@ -88,10 +86,47 @@ func (h *Handlers) ExtendVoteHandler() sdk.ExtendVoteHandler {
...
+				h.logger.Error("unexpected error while checking previous validator tree entry", "err", err)
+			}
+			return nil, err
+		}
+
+		val, err := h.stakingKeeper.GetValidator(ctx, h.signer.GetValAddress())
+		if err != nil {
+			return nil, err
+		}
+		valKeys, err := h.pubKeyKeeper.GetValidatorKeys(ctx, val.OperatorAddress)
```
Thomasvdam, line 116: Is the address we get from `h.signer.GetValAddress` not already the operator address? In my seda_keys.json I see `validator_addr: sedavaloper1ymuuw4cfa5j60hdecymfjkhsfhvxtt0ujfknvx`

app/utils/seda_signer.go
```diff
@@ -0,0 +1,133 @@
...
+
+// IsLoaded returns true if the signer is loaded and ready for signing.
+func (s *sedaKeys) IsLoaded() bool {
+	return s.isLoaded
+}
+
+// Reload reloads the signer from the key file.
+func (s *sedaKeys) reload() error {
+	keys, err := loadSEDAKeys(s.keyPath)
+	if err != nil {
```
Thomasvdam, line 124: Maybe we should set reset the state of the signer when loading fails?

Thomasvdam (approved):

app/abci/handlers.go
hacheigriega: addressed in https://github.com/sedaprotocol/seda-chain/pull/413/commits/885432acd25f3dfbeaacb601da15dcdc8dbf5698

app/utils/seda_signer.go
hacheigriega: addressed in https://github.com/sedaprotocol/seda-chain/pull/413/commits/885432acd25f3dfbeaacb601da15dcdc8dbf5698

PR #416, opened 2024-11-12
refactor(batching): Batch query refactor

hacheigriega:
## Explanation of Changes

- For query of all batches, do not return batches without signatures.


Thomasvdam (approved):
I think it'd be nice if we add a query flag (if that's even possible :')) to also include batches without signatures. This might be interesting for people who are building optimised solvers so they can start building the proofs and send right away when the signatures come in.

Definitely a separate issue kind of thing though!

hacheigriega:
@Thomasvdam interesting idea! That should be simple to implement. I'll make sure to address that. There are a few other query/store refactors that need be addressed. For instance, the way we store the tree entries right now does not accommodate for multiple batching keys. I'll create another PR for those then.

Issue #421, opened 2024-11-25
✨ Implement a state for indicating if proving scheme is enabled


hacheigriega:
## ✨ Feature

This issue calls for a mechanism for checking if a proving scheme is enabled and storing the boolean state in `x/pubkey`. 

- Inside the `x/pubkey` end block, check the proportion of validators that have registered a key corresponding to a given proving scheme. 
- There will be a boolean state that indicates if a proving scheme is enabled. This state will default to `false` until the end block check passes. This state will be used by:
    - vote extension handlers to check which proving schemes are active
    - create-validator and unjail transactions as described in #422

PR #427, opened 2024-11-29
Add proving schemes state to pubkey module

hacheigriega:
## Explanation of Changes

The public key end blocker now checks the percentage of validators that have registered the secp256k1 public key on chain. If this check passes the 80% threshold, the activation height of the proving scheme is set to `ActivationLag` blocks after the block at which the threshold is reached. When the activation height is reached, the proving scheme is set to active status and all validators without the required public key gets jailed.

This activation height is set to -1 by default to signify that activation process is not in progress. Also note that if the public key registration rate goes below the 80% threshold after an activation process begins (due to a new validator without the required key joining with substantial voting power), the activation height is reverted to -1. 

For testing convenience, this `ActivationLag` is currently set to a small number, 25. This value is a parameter that is subject to change through governance process.

Batching is skipped until the secp256k1 proving scheme is activated.

## Testing
Added simple unit tests but jailing is not tested yet.
Tested activation process & cancellation of activation process on a local multi-node network.

Closes: #421


x/pubkey/keeper/endblock.go
```diff
@@ -0,0 +1,76 @@
...
+	if err != nil {
+		return err
+	}
+	if isEnabled {
+		return
+	}
+
+	// If the sum of the voting power has reached 80%, enable secp256k1
+	// proving scheme.
+	totalPower, err := k.stakingKeeper.GetLastTotalPower(ctx)
```
Thomasvdam, line 39: We might want to add a delay here so validators have a chance not to get jailed. So once we hit 80% we wait X blocks before enabling it.

x/pubkey/keeper/keeper.go
```diff
@@ -88,3 +94,94 @@ func (k Keeper) GetValidatorKeys(ctx context.Context, validatorAddr string) (res
...
+	for ; itr.Valid(); itr.Next() {
+		kv, err := itr.KeyValue()
+		if err != nil {
+			return nil, err
+		}
+
+		// Skip if the validator has already been processed.
+		if bytes.Equal(kv.Key.K1(), currentVal) {
+			continue
+		}
```
Thomasvdam, line 118: ~~Why do we need this check?~~

Ooh, I guess the iterator is over the entire store, so with multiple proving schemes we would get multiple hits for the same validator. And since the iterator is 'sorted' we can simply look at the previous iterations value?

x/pubkey/keeper/endblock.go
hacheigriega: Added activation height to the proving scheme type (added explanation in the prompt)

x/pubkey/keeper/keeper.go
hacheigriega: Yes exactly. I suppose I could add a prefix iterator inside the for loop. Will look into this tmrw.

hacheigriega: Hm, It doesn't look like there is a better way to iterate over K1s.

scripts/local_multi_setup.sh
```diff
@@ -199,9 +199,9 @@ echo "4 validators are up and running!"
...
-$BIN tx pubkey add-seda-keys --from validator4 --keyring-backend test --home $HOME/.sedad/validator4 --gas-prices 10000000000aseda --gas auto --gas-adjustment 2.0 --keyring-backend test --chain-id=testing --node http://localhost:26657 --yes
+# $BIN tx pubkey add-seda-keys --from validator2 --keyring-backend test --home $HOME/.sedad/validator2 --gas-prices 10000000000aseda --gas auto --gas-adjustment 2.0 --keyring-backend test --chain-id=testing --node http://localhost:26657 --yes
+# $BIN tx pubkey add-seda-keys --from validator3 --keyring-backend test --home $HOME/.sedad/validator3 --gas-prices 10000000000aseda --gas auto --gas-adjustment 2.0 --keyring-backend test --chain-id=testing --node http://localhost:26657 --yes
+# $BIN tx pubkey add-seda-keys --from validator4 --keyring-backend test --home $HOME/.sedad/validator4 --gas-prices 10000000000aseda --gas auto --gas-adjustment 2.0 --keyring-backend test --chain-id=testing --node http://localhost:26657 --yes
```
Thomasvdam, line 204: This was just for testing right?

x/pubkey/keeper/endblock.go
```diff
@@ -0,0 +1,155 @@
...
+			return err
+		}
+		_, err = k.GetValidatorKeyAtIndex(ctx, valAddr, keyIndex)
+		if err != nil {
+			if errors.Is(err, collections.ErrNotFound) {
+				consAddr, err := val.GetConsAddr()
+				if err != nil {
+					return err
+				}
+				err = k.slashingKeeper.Jail(ctx, consAddr)
```
Thomasvdam, line 139: Is there a reason we're jailing through the slashing module and not the staking module?

x/pubkey/keeper/endblock_test.go
```diff
@@ -0,0 +1,246 @@
...
+}
+
+func TestEndBlock(t *testing.T) {
+	f := initFixture(t)
+	ctx := f.Context()
+
+	f.keeper.InitGenesis(ctx, *types.DefaultGenesisState())
+
+	_, valAddrs, _ := createValidators(t, f, []int64{1, 3, 5, 7, 2, 1}) // 1+3+5+7+2+1 = 19
+	pubKeys := generatePubKeys(t, 6)
```
Thomasvdam, line 207: Should we not leave 1 validator out so we can test the jailing?

x/pubkey/keeper/endblock.go
hacheigriega: No big reason. I just saw that the evidence module calls the slashing module one, which wraps the staking module jail and additionally emits an event.

scripts/local_multi_setup.sh
hacheigriega: oops thanks

Thomasvdam (approved):

proto/sedachain/pubkey/v1/genesis.proto
```diff
@@ -11,6 +11,8 @@ option go_package = "github.com/sedaprotocol/seda-chain/x/pubkey/types";
 message GenesisState {
   repeated ValidatorPubKeys validator_pub_keys = 1
       [ (gogoproto.nullable) = false ];
+  repeated ProvingScheme proving_schemes = 2 [ (gogoproto.nullable) = false ];
+  Params params = 3 [ (gogoproto.nullable) = false ];
```
mariocao, line 15: I know that adding new fields should be always at the end to make it not a breaking change. But does it make sense to consider having Params at index 1 as most modules in Cosmos SDK?

x/pubkey/keeper/endblock.go
```diff
@@ -0,0 +1,155 @@
...
+		//nolint:gosec // G115: We shouldn't get negative power anyway.
+		powerSum += uint64(power)
+		return false
+	})
+	if err != nil {
+		return false, err
+	}
+
+	//nolint:gosec // G115: We shouldn't get negative power anyway.
+	requiredPower := uint64(totalPower.Int64()*100*4/5 + 1)
```
mariocao, line 108: Should we have this 80% be a governance parameter?
The only constraint I see it needs to have is that it needs to be bigger than 2/3, right?

mariocao: Or at least a constant somewhere 😅

x/pubkey/types/params.go
```diff
@@ -0,0 +1,24 @@
...
+)
+
+const (
+	DefaultActivationLag = 25
+)
+
+// DefaultParams returns default pubkey module parameters.
+func DefaultParams() Params {
+	return Params{
+		ActivationLag: DefaultActivationLag,
```
mariocao, line 14: Maybe delay is more explicit. Wdyt?
I would suggest `activationBlockDelay` or `activationDelay` (but then in comment that it is in blocks and not time).

```suggestion
		ActivationBlockDelay: DefaultActivationBlockDelay,
```

mariocao (commented):
Apart from my small few comments, everything looks great! 🥇 🙌

x/pubkey/types/params.go
```diff
@@ -5,20 +5,25 @@ import (
...
 
 // ValidateBasic performs basic validation on pubkey module parameters.
 func (p *Params) Validate() error {
-	if p.ActivationLag < 0 {
-		return sdkerrors.ErrInvalidRequest.Wrapf("ActivationLag should not be negative: %d", p.ActivationLag)
+	if p.ActivationBlockDelay < 0 {
+		return sdkerrors.ErrInvalidRequest.Wrapf("ActivationBlockDelay should not be negative: %d", p.ActivationBlockDelay)
+	}
+	if p.ActivationThresholdPercent < 66 || p.ActivationThresholdPercent > 100 {
+		return sdkerrors.ErrInvalidRequest.Wrapf("ActivationThresholdPercent should be between 66 and 100: %d", p.ActivationThresholdPercent)
```
hacheigriega, line 26: 67, right? 😅

x/pubkey/keeper/endblock.go
```diff
@@ -104,8 +104,12 @@ func (k Keeper) CheckKeyRegistrationRate(ctx sdk.Context, keyIndex utils.SEDAKey
...
 	}
 
-	//nolint:gosec // G115: We shouldn't get negative power anyway.
-	requiredPower := uint64(totalPower.Int64()*100*4/5 + 1)
+	activationThresholdPercent, err := k.GetActivationThresholdPercent(ctx)
+	if err != nil {
+		return false, err
+	}
+
+	requiredPower := totalPower.Uint64()*uint64(activationThresholdPercent) + 1
```
hacheigriega, line 112: I suppose we can just store `activation_threshold_percent` as uint64?

PR #428, opened 2024-12-02
Make dr block height part of the primary key

Thomasvdam:
## Motivation

This allows us to keep old results around for recurring data request IDs.

## Explanation of Changes

We don't need a latest batch assignment query since those should always uniquely identify the exact DR_ID+height combination for which the result has been batched.

We have to manually create a range since the built in `collections.NewSuperPrefixedTripleRange` doesn't expose the ordering methods despite them being present on the struct. :( Ver sad.

Also introduces a few missing genesis states during initialisation.

## Testing

Expanded on the result inclusion test to verify that you get the latest result when you do not specify the height, but by specifying the height you can still retrieve an older result. This also brought up the issue that by default the ordering of a Range is ascending.

## Related PRs and Issues

N.A.


x/batching/keeper/data_result.go
```diff
@@ -50,60 +50,118 @@ func (k Keeper) GetDataResult(ctx context.Context, dataReqID string) (*types.Dat
...
+// data request's ID.
+func (k Keeper) GetLatestDataResult(ctx context.Context, dataReqID string) (*types.DataResult, error) {
+	dataResult, err := k.getLatestDataResult(ctx, false, dataReqID)
+	if err != nil {
+		if errors.Is(err, collections.ErrNotFound) {
+			// Look among batched data requests.
+			dataResult, err := k.getLatestDataResult(ctx, true, dataReqID)
+			if err != nil {
+				if errors.Is(err, collections.ErrNotFound) {
+					return nil, nil
```
hacheigriega, line 63: Not returning an error here could cause a nil dereference error in the querier, right?

x/batching/keeper/data_result.go
```diff
@@ -12,31 +12,31 @@ import (
...
+func (k Keeper) GetDataResult(ctx context.Context, dataReqID string, dataReqHeight uint64) (*types.DataResult, error) {
+	dataResult, err := k.dataResults.Get(ctx, collections.Join3(false, dataReqID, dataReqHeight))
 	if err != nil {
 		if errors.Is(err, collections.ErrNotFound) {
 			// Look among batched data requests.
-			dataResult, err := k.dataResults.Get(ctx, collections.Join(true, dataReqID))
+			dataResult, err := k.dataResults.Get(ctx, collections.Join3(true, dataReqID, dataReqHeight))
 			if err != nil {
 				if errors.Is(err, collections.ErrNotFound) {
 					return nil, nil
```
hacheigriega, line 42: Same issue here. Perhaps we shouldn't catch the `ErrNotFound` error? I'm also worried about inconsistencies in dealing with this error since most of our getters don't catch this error

x/batching/keeper/data_result.go
Thomasvdam: Good catch. I guess adding a null check in the querier makes the most sense?

hacheigriega (approved):

Issue #425, opened 2024-11-25
✨ Update tally gas computations


Thomasvdam:
## ✨ Feature

With the introduction of the `tally_gas_limit` field on the data request we need to update how we forward the gas limit to the VM, as well as the computation for the total gas used. Related to #424 

## Motivation

More granular error messaging when a DR runs out of gas.

## Implementation

- [x] Use `tally_gas_limit` as limit for the tally VM
- [x] Compute gas used as the sum of all `gas_used` values from the submitted reveals + the gas used by the tally VM.
- [x] Clamp max gas to a sensible value before passing it to the VM (300 TGas? look at CosmWasm/Near/Executor)
- [x] Update env var name: `DR_GAS_LIMIT` -> `DR_TALLY_GAS_LIMIT`
- [x] Add more env variables for the tally VM.

PR #432, opened 2024-12-05
Tally gas metering

Thomasvdam:
## Motivation

Keep track of the gas consumed so we can implement rewards and burn later.

## Explanation of Changes

Update the SEDA tallyvm to v2.1.1. Move the reveal ordering logic to the
    top of the FilterAndTally method so we can always calculate the
    execution gas used, even if something else fails.

Also update remove the authority address from the batching module as this isn't used at the moment.

## Testing

Updated unit and integration tests.

## Related PRs and Issues

Closes: #425


app/app.go
```diff
@@ -666,7 +665,7 @@ func NewApp(
 	app.BatchingKeeper = batchingkeeper.NewKeeper(
 		appCodec,
 		runtime.NewKVStoreService(keys[batchingtypes.StoreKey]),
-		authtypes.NewModuleAddress(batchingtypes.ModuleName).String(),
+		authtypes.NewModuleAddress(govtypes.ModuleName).String(),
```
hacheigriega, line 668: Hm it doesn't look like this is used anywhere since batching doesn't have update params tx. Should we add a quick issue for it?

Thomasvdam: Oh shit 😂 Yeah probably better to remove it until we actually need it

hacheigriega (approved):

PR #433, opened 2024-12-07
feat: params query

gluax:
## Motivation

Seems that most built in modules support querying the params from GRPC/CLI.

This would make my life much easier on the test suite side as well.

## Explanation of Changes

I added a new `Params` query to the pubkey module, matching how other sdk modules add the `params` query.
I also added the CLI/Go code to support the above.

As for the other protobuf changes, I'm not sure why the `make proto-all` command was removing the 1 line from a few files.

Currently branch is based on `main`, if it's a pain I can re-base and target #431 instead @hacheigriega?

## Testing

Tested with test suite.

image

Should I somehow add other tests?

## Related PRs and Issues

N/A


proto/sedachain/pubkey/v1/query.proto
```diff
@@ -11,6 +11,11 @@ option go_package = "github.com/sedaprotocol/seda-chain/x/pubkey/types";
 
 // Query defines the gRPC querier service.
 service Query {
+  // Params returns the total set of pubkey parameters.
+  rpc Params(QueryParamsRequest) returns (QueryParamsResponse) {
+    option (google.api.http).get = "/cosmos/mint/params";
```
Thomasvdam, line 16: This should probably be `"/sedachain/pubkey/params";`

Thomasvdam (commented):
> As for the other protobuf changes, I'm not sure why the make proto-all command was removing the 1 line from a few files.

Last time we had this is was because we were using different versions of the proto toolchain. I'm using buf `1.47.2`

proto/sedachain/pubkey/v1/query.proto
gluax: Whoops forgot to change that when looking at how Cosmos did this my bad 😅

gluax:
> > As for the other protobuf changes, I'm not sure why the make proto-all command was removing the 1 line from a few files.
> 
> Last time we had this is was because we were using different versions of the proto toolchain. I'm using buf `1.47.2`

@Thomasvdam I updated my buf version to match yours, but it seems those 1 line changes remain the same.

Thomasvdam:
> > > As for the other protobuf changes, I'm not sure why the make proto-all command was removing the 1 line from a few files.
> > 
> > 
> > Last time we had this is was because we were using different versions of the proto toolchain. I'm using buf `1.47.2`
> 
> @Thomasvdam I updated my buf version to match yours, but it seems those 1 line changes remain the same.

That's odd, when I run it on your branch it adds all the lines back. Are you sure the shell session executing the `proto-gen` has the correct buf version in its path?

gluax:
> > > > As for the other protobuf changes, I'm not sure why the make proto-all command was removing the 1 line from a few files.
> > > 
> > > 
> > > Last time we had this is was because we were using different versions of the proto toolchain. I'm using buf `1.47.2`
> > 
> > 
> > @Thomasvdam I updated my buf version to match yours, but it seems those 1 line changes remain the same.
> 
> That's odd, when I run it on your branch it adds all the lines back. Are you sure the shell session executing the `proto-gen` has the correct buf version in its path?

image


I'm using the `make proto-all` command to generate the protobuf, is that different from what you're using?

gluax:
@Thomasvdam and I learned I had to do `make proto-dep-install` first.

Thomasvdam (approved):
Whoops

PR #439, opened 2024-12-11
Improve tally error handling

### LINKED ISSUE #430: 🔧 Missing tally oracle program error is not too clear
## 🔧 Chore

The error written to the result of a DR that specifies a missing tally oracle program is a little unclear:

```
collections: not found: key 'hexBytes:67268f7243c5e50084da85b8fec4934047715c3b35393a9dbd850f1e3fc9e2c6' of type github.com/cosmos/gogoproto/sedachain.wasm_storage.v1.OracleProgram
```

We should be more descriptive and leave out unnecessary implementation details.

## Motivation

Make it easier for developers to see what went wrong.

hacheigriega:
## Explanation of Changes

When an error occurs in FilterAndTally(), we now log the error details for debugging purposes and return simplified error information to be stored as part of the data results.

## Related PRs and Issues

Closes #430

### DIFF
--- a/x/tally/keeper/abci.go
+++ b/x/tally/keeper/abci.go
@@ -9,8 +9,6 @@ import (
 	"strconv"
 	"strings"
 
-	errorsmod "cosmossdk.io/errors"
-
 	sdk "github.com/cosmos/cosmos-sdk/types"
 
 	"github.com/sedaprotocol/seda-wasm-vm/tallyvm/v2"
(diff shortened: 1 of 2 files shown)


x/tally/keeper/abci.go
```diff
@@ -117,6 +113,9 @@ func (k Keeper) ProcessTallies(ctx sdk.Context, coreContract sdk.AccAddress) err
 				dataResults[i].ExitCode = uint32(result.exitInfo.ExitCode)
 				dataResults[i].Result = result.result
 			}
+			dataResults[i].Consensus = result.consensus
+			dataResults[i].GasUsed = result.execGasUsed + result.tallyGasUsed
```
Thomasvdam, line 117: We should always set the gas used. If the requestor specified a wrong tally program or incorrect filter they still should pat for the computation that was used for their request.

hacheigriega: I just thought it makes more sense to check the value of `err` right after its assignment as it's always done in golang.

Thomasvdam: Aaah, on mobile it looked like it was moved into the `else` clause. Never mind.

Thomasvdam (approved):

mariocao (approved):

PR #441, opened 2024-12-11
Publish correct tally completion event values

Thomasvdam:
## Motivation

Allow indexing the tally logs and gas used attributes.

## Explanation of Changes

Turns out we were accidentally shadowing the `result` variable in the switch statement, which meant that when we assigned result after the statment we used the empty reference with default values.

## Testing

Verified that a local node with these changes output the correct events at block #9777 with the (at the time of writing) current planet.

## Related PRs and Issues

N.A.


mariocao (approved):

Thomasvdam:
Fat finger, apologies.

hacheigriega (approved):
wow, time to start checking events

PR #442, opened 2024-12-11
Support amino signing for data proxy transactions

Thomasvdam:
## Motivation

We want to support data proxy operators utilising hardware wallets (Ledger) for managing a data proxy. 

## Explanation of Changes

I'm not 100% certain on all the proto amino tags. I looked at various cosmos SDK modules to see how those were implemented and then it was a bunch of trial and error until I got all transactions to pass locally. 🤷 

If there's any good sources on what proto annotations we need exactly I'd love to read more!

## Testing

Fairly involved local setup where I ran the chain locally using `seda-1-planet` as the chain ID, enabled CORS for all domains in the config.toml, set up the indexer and explorer locally (and all related services), and used a HTTPS proxy (proxyman) to route traffic from https://rpc.planet.seda.xyz to the local RPC port.

With this entire setup I was able to use a local build of the explorer to sign TXs with amino sign mode. I tried registering a data proxy, editing a single field at a time (payoutAddress, memo, fee, fee+update delay), edit multiple fields at a time, and transferring the admin address. The web code still needs to be cleaned up though, will open that PR tomorrow. 

## Related PRs and Issues

Part-of: https://github.com/sedaprotocol/seda-explorer/issues/591


hacheigriega (approved):
I am hoping to understand this test setup soon.. Seems very neat.
Is it OK that our other txs don't support amino signing?

mariocao (approved):

Thomasvdam:
> I am hoping to understand this test setup soon.. Seems very neat. Is it OK that our other txs don't support amino signing?

We should definitely test the new create validator and unjail transactions. For the rest I don't think it's urgent, but worth keeping in the back of our minds. :)

PR #450, opened 2024-12-17
Gas metering for filters

### LINKED ISSUE #449: ✨ Consensus filter gas consumption
## ✨ Feature

We want the tally consensus filter to also incur a gas cost and count towards the `tally_gas_limit` of the data request. For now we can go with a constant that we multiply with the replication factor (except in the `NONE` filter case). We're not (yet) considering the filter size.

## Motivation

As it's computation that's part of the SEDA protocol it makes sense that requestors fund this computation done by the validators.

## Implementation

- Add state for a constant gas cost for the `NONE` filter.
- Add state for a constant gas cost multiplier for the `MODE` filter.
- Add state for a constant gas cost multiplier for the `STD_DEV` filter.
- Apply gas cost in the tally after executing the filter. `MODE` and `STD_DEV` costs can simply be multiplied with the replication factor.
- Add parameters (and default values 😬) for the supported filters.

hacheigriega:
## Explanation of Changes

Adding gas metering for filters as described in Issue #449.

The gas used amount is constant for a given filter type and replication factor regardless of whether the filter returned an error or not. If the filter was not applied at all, the gas used is zero. 

The gas used towards filtering is subtracted from the gas limit we feed into the tally VM. 
In addition, the tally gas used amount is now the sum of the gas used by the tally VM and the gas used by the filter.

## Testing

Added checks to the existing filter tests

## Related PRs and Issues

Closes #449


x/tally/keeper/filter.go
```diff
@@ -39,34 +50,46 @@ func ApplyFilter(input []byte, reveals []types.RevealBody) ([]int, bool, []strin
...
 
 	var filter types.Filter
-	var err error
 	switch input[0] {
 	case filterTypeNone:
 		filter, err = types.NewFilterNone(input)
+		result.GasUsed = params.FilterGasCostNone
 	case filterTypeMode:
 		filter, err = types.NewFilterMode(input)
+		result.GasUsed = params.FilterGasCostMultiplierMode * uint64(replicationFactor)
```
Thomasvdam, line 69: Is there a reason the replication factor is an `int64`? On the Rust side it's a `u16`.

x/tally/keeper/endblock.go
```diff
@@ -247,7 +248,7 @@ func (k Keeper) FilterAndTally(ctx sdk.Context, req types.Request) (TallyResult,
 	if err != nil {
 		return result, k.logErrAndRet(ctx, err, types.ErrGettingMaxTallyGasLimit, req)
 	}
-	gasLimit := min(req.TallyGasLimit, maxGasLimit)
+	gasLimit := min(req.TallyGasLimit, maxGasLimit) - filterResult.GasUsed
```
Thomasvdam, line 251: We should verify that the `filterResult.GasUsed` is larger than the `gasLimit` to prevent insanely high gas limits :p

x/tally/keeper/filter.go
hacheigriega: ah didn't realize it. updated to `uint16`

x/tally/keeper/endblock.go
hacheigriega: changed it to 
```
gasLimit := max(min(req.TallyGasLimit, maxGasLimit)-filterResult.GasUsed, 0)
```

Thomasvdam: That won't do anything since they're unsigned integers right? 0 - 100 = 18446744073709551516 in uint64 math.

x/tally/keeper/integration_helpers_test.go
```diff
@@ -0,0 +1,287 @@
...
+import (
+	"encoding/base64"
+	"encoding/binary"
+	"encoding/hex"
+	"encoding/json"
+	"fmt"
+	"testing"
+
+	"github.com/cometbft/cometbft/crypto/secp256k1"
+	vrf "github.com/sedaprotocol/vrf-go"
```
mariocao, line 12: This is old, right?
I don't think we should be adding any VRF library (yet).

x/tally/keeper/integration_helpers_test.go
```diff
@@ -0,0 +1,287 @@
...
+	allBytes = append(allBytes, memoHash...)
+	allBytes = append(allBytes, chainIDBytes...)
+	allBytes = append(allBytes, contractAddrBytes...)
+	allBytes = append(allBytes, sequenceBytes...)
+
+	hasher.Reset()
+	hasher.Write(allBytes)
+	hash := hasher.Sum(nil)
+
+	proof, err := vrf.NewK256VRF().Prove(signKey, hash)
```
mariocao, line 220: If not mistaken, currently we are using a Hash of the Reveal Body.

x/tally/keeper/integration_helpers_test.go
hacheigriega: I just wanted to replicate the overlay code and saw that overlay uses vrf library to generate the stake, commit, and reveal proofs.

x/tally/keeper/integration_helpers_test.go
hacheigriega: For commitments, right? I saw that the contract additionally requires proofs for stake, commit, and reveal txs.

x/tally/keeper/endblock.go
hacheigriega: duh.. thanks again. fixed it to
```
	var gasLimit uint64
	if min(req.TallyGasLimit, maxGasLimit) > filterResult.GasUsed {
		gasLimit = min(req.TallyGasLimit, maxGasLimit) - filterResult.GasUsed
	} else {
		gasLimit = 0
	}
```

x/tally/keeper/integration_helpers_test.go
mariocao: yep yep!! 👍

mariocao (approved):

PR #452, opened 2024-12-20
Refactor tally process to updated specs

### LINKED ISSUE #451: ♻️ Tally refactor to updated specs
## ♻️ Refactor

Before implementing incentive features for tally, refactor the tally process to the updated specs described [here](https://www.notion.so/sedaprotocol/x-tally-161a68d575ca8032ad1fd4714a48ec4b?pvs=4).

Specific tasks would involve:

- [x] Reordering the tally process to follow: filtering -> tally execution -> payout calculation
- [x] Requestor-provided filter should be fully validated before the specific filter is applied. In other words, any errors due to requestor's fault in constructing the filter input should be handled before the specific filter is applied.
- [x] Update the tally process logic. ~~Consensus is now determined by tracking lists of successful reports and error reports, where a successful report is a zero-exit code reveal in "agreement" and an error report is any non-zero exit code reveal or a reveal that is not in "agreement."~~
- [x] Update the relevant tests accordingly. Add more tests if necessary.

hacheigriega:
## Motivation

Refactor tally process to match the updated specs. 

The biggest change in terms of the overall tally flow is that whether we execute tally or not no longer depends on the consensus determined by the filtering process. The filter now returns one of the following states, and tally is executed only in the last case:

- Consensus in error: More than 2/3 of reveals contain non-zero exit codes or cannot be parsed.
- No consensus: There is no consensus either in error or in data.
- Consensus in data (no error): Consensus has been reached in reveal data. Tally is executed.

## Explanation of Changes

Consensus in error is determined using `errors`, a slice of boolean indicating whether i-th reveal contains an error (either non-zero exit code or parse error). This slice is passed around during the filtering phase to be updated whenever an error is detected in a reveal. 

The specific filters (none, mode, and standard deviation) have been refactored to fully separate out the construction part. This way, the data requestor-provided tally input is fully validated before the filter execution.

## Related PRs and Issues

Closes #451


x/tally/types/filters.go
```diff
@@ -99,16 +119,19 @@ func (f FilterMode) ApplyFilter(reveals []RevealBody) ([]int, error) {
 
 type FilterStdDev struct {
 	maxSigma   Sigma
-	numberType byte
 	dataPath   string // JSON path to reveal data
+	filterFunc stdDevFilterFunc
+	gasCost    uint64
 }
 
+type stdDevFilterFunc func(dataList []any, maxSigma Sigma) ([]int, error)
```
hacheigriega, line 127: note to self: no need to declare this type

x/tally/keeper/endblock.go
```diff
@@ -94,11 +94,11 @@ func (k Keeper) ProcessTallies(ctx sdk.Context, coreContract sdk.AccAddress) err
 		}
 
 		switch {
-		case len(req.Commits) < int(req.ReplicationFactor):
+		case len(req.Commits) == 0 || len(req.Commits) < int(req.ReplicationFactor):
 			dataResults[i].Result = []byte(fmt.Sprintf("need %d commits; received %d", req.ReplicationFactor, len(req.Commits)))
 			dataResults[i].ExitCode = batchingtypes.TallyExitCodeNotEnoughCommits
 			k.Logger(ctx).Info("data request's number of commits did not meet replication factor", "request_id", req.ID)
-		case len(req.Reveals) < int(req.ReplicationFactor):
+		case len(req.Reveals) == 0 || len(req.Reveals) < int(req.ReplicationFactor):
```
Thomasvdam, line 101: I think we should handle this case differently, as receiving 9/10 reveals can still potentially lead to a result where 7 answers are the same (or considered similar enough when using deviation). I guess that does mean we have to pass the reveals and RF separately, or we use a `nil` value to signal a missing reveal.

x/tally/keeper/filter.go
```diff
@@ -28,69 +71,42 @@ type FilterResult struct {
...
 		if freq[tuple] > maxFreq {
-			proxyPubKeys = reveal.ProxyPubKeys
+			result.ProxyPubKeys = reveal.ProxyPubKeys
 			maxFreq = freq[tuple]
 		}
 	}
-	if maxFreq*3 < len(reveals)*2 {
+	if maxFreq*3 < len(reveals)*2 { // TODO remove basic consensus check?
 		return result, types.ErrNoBasicConsensus
 	}
```
Thomasvdam, line 96: I think it makes sense to keep this short-circuit in. Makes the rest of the filter code much easier to read.

x/tally/keeper/endblock.go
```diff
@@ -209,44 +209,57 @@ func (k Keeper) FilterAndTally(ctx sdk.Context, req types.Request) (TallyResult,
...
 	}
-	paybackAddrHex := hex.EncodeToString(decodedBytes)
-
-	filterResult, err := k.ApplyFilter(ctx, filter, reveals, req.ReplicationFactor)
+	filterResult, err := ApplyFilter(filter, reveals)
 	result.consensus = filterResult.Consensus
 	result.proxyPubKeys = filterResult.ProxyPubKeys
 	if err != nil {
 		return result, k.logErrAndRet(ctx, err, types.ErrApplyingFilter, req)
 	}
```
Thomasvdam, line 222: This is not correct, as we should still calculate the exec gas used and determine the payouts in the case of consensus on a non-zero exit code.

x/tally/keeper/filter.go
```diff
@@ -28,69 +71,42 @@ type FilterResult struct {
...
-	}
+	outliers, consensus := filter.ApplyFilter(reveals, result.Errors)
 
-	outliers, err := filter.ApplyFilter(reveals)
 	switch {
-	case err == nil:
-		result.Outliers = outliers
+	case errorCount(result.Errors)*3 > len(reveals)*2:
+		result.Consensus = true
+		return result, types.ErrConsensusInError
```
Thomasvdam, line 103: We could have outliers here right? And I think marking it as an error makes it more difficult to reason about at the callsite as this error and `ErrNoConsensus` will still have to go through phase III (determining gas used and payouts).

x/tally/keeper/filter.go
```diff
@@ -16,10 +16,53 @@ const (
 )
 
 type FilterResult struct {
-	Outliers     []int    // outlier list
-	Consensus    bool     // whether consensus was reached
-	ProxyPubKeys []string // consensus data proxy public keys
-	GasUsed      uint64   // gas used for filter
+	Errors       []bool   // i-th item is true if i-th reveal is non-zero exit or corrupt
+	Outliers     []int    // i-th item is non-zero if i-th reveal is an outlier
```
Thomasvdam, line 20: Maybe it makes sense to have it as a `[]bool` here as well and only convert it to a `[]int` when we pass it to the tallyvm? Feels a bit odd to have that implementation detail leak into the filters here.

x/tally/types/filters.go
```diff
@@ -8,52 +8,51 @@ import (
...
+	gasCost  uint64
 }
 
 // NewFilterMode constructs a new FilerMode object given a filter
 // input.
 // Mode filter input looks as follows:
 // 0             1                  9       9+data_path_length
 // | filter_type | data_path_length |   data_path   |
-func NewFilterMode(input []byte) (FilterMode, error) {
+func NewFilterMode(input []byte, gasCostMultiplier uint64, replicationFactor uint16) (FilterMode, error) {
```
Thomasvdam, line 55: We could consider storing the replication factor on the filter, since we're passing that data anyway to determine the gas cost. That way we can 'simply' refer to the replication factor wherever we did `len(reveals)` and that should also simplify handling missing reveals? 🤔

x/tally/keeper/filter.go
hacheigriega: My reasoning was that we return an error here if we do not want tally VM execution. 
I forgot that we need outlier list when we need to calculate the payouts. Added.

x/tally/keeper/endblock.go
hacheigriega: Ah right, thanks. I refactored the way we handle errors in this function. For that matter though, we ALWAYS want to calculate exec gas & payouts right? That is, we go through phase III even when there is an error while building filter or when there is no consensus.

x/tally/keeper/filter.go
hacheigriega: Addressed in [bc1f126](https://github.com/sedaprotocol/seda-chain/pull/452/commits/bc1f12662c403d401343f9e894e487768d8ddc55). Actually, tally vm execution testing passed even though I didn't convert it to integers when passing it as an argument. Would it be ok to just pass the outlier list argument as JSON array of booleans?

x/tally/keeper/endblock.go
hacheigriega: Good idea and I think it makes sense to address this in a separate PR together with issue #453 ?

hacheigriega:
@Thomasvdam Ready for re-review now whenever you have time! Thank you!!

x/tally/types/filters.go
hacheigriega: I will address this in a separate PR as a part of #453 if that makes sense.

x/tally/keeper/endblock.go
Thomasvdam: > That is, we go through phase III even when there is an error while building filter or when there is no consensus.

Yes, not sure how/if we do exec gas calculations though, but there should be some sort of payouts I guess. Gets even more fun if we still want to pay out data proxies :P

x/tally/keeper/endblock.go
```diff
@@ -210,93 +201,39 @@ func (k Keeper) FilterAndTally(ctx sdk.Context, req types.Request) (TallyResult,
...
+		result.proxyPubKeys = filterResult.ProxyPubKeys
 
-	// Phase II: Tally Program Execution
-	vmRes, err := k.ExecuteTallyProgram(ctx, req, filterResult, reveals)
-	if err != nil {
-		return TallyResult{}, err
+		// Phase II: Tally Program Execution
+		if err != nil {
+			result.result = []byte(err.Error())
+			result.exitInfo.ExitCode = TallyExitCodeFilterError
```
Thomasvdam, line 217: We still consumed the filter gas here technically? 🤔

x/tally/keeper/endblock.go
```diff
@@ -210,93 +201,39 @@ func (k Keeper) FilterAndTally(ctx sdk.Context, req types.Request) (TallyResult,
...
+		} else {
+			vmRes, err := k.ExecuteTallyProgram(ctx, req, filterResult, reveals)
+			if err != nil {
+				result.result = []byte(err.Error())
+				result.exitInfo.ExitCode = TallyExitCodeFilterError
```
Thomasvdam, line 222: Same here. Not sure if there's a scenario where the tally consumes gas but still returns an error, but just to be sure we should probably set it similar to the success case.

x/tally/keeper/filter.go
gluax: Where are these outliers being passed to? the tallyvm? Or somewhere else?

hacheigriega: Outlier list is one of the arguments. It is now a JSON array of booleans rather than of ints. I forget why I used ints initially, but I think it should be up to the wasm program to parse the outlier list correctly.

x/tally/keeper/filter.go
```diff
@@ -28,69 +71,44 @@ type FilterResult struct {
...
-		return result, err
-	}
+	outliers, consensus := filter.ApplyFilter(reveals, result.Errors)
 
-	outliers, err := filter.ApplyFilter(reveals)
 	switch {
-	case err == nil:
+	case errorCount(result.Errors)*3 > len(reveals)*2:
+		result.Consensus = true
 		result.Outliers = outliers
```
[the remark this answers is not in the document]
Thomasvdam: I think for this PR this is ok, but we'll probably need another pass once we have everything crystal clear in terms of short circuits, edge cases, etc.

Thomasvdam (approved):

x/tally/keeper/filter.go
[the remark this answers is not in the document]
hacheigriega: Sounds good and I agree. We need to make sure the tests cover all edge cases. This issue was actually caught by a test, but I mistakenly ignored it 😢

PR #461, opened 2025-01-10
Payout to committers in case of data request timeout

### LINKED ISSUE #453: ✨ Incur gas cost for data requests that time out
## ✨ Feature

We want a small portion of the funds attached to a timed-out data request to be distributed to the executors that did commit.

## Motivation

This makes it less risky for the solver to escrow their funds, incentives requestors to properly price their data requests, and makes it less risky for an executor node to post a commitment.

## Implementation

- [x] Add param for a fixed amount of gas that will be charged per commitment in the case of a timeout. This amount should at least be equal to the gas cost of spinning up a VM.
- [x] When a DR times out due to not receiving enough commitments calculate the gas used based on the param.
- [x] Create the correct distribution messages when removing the DR from the contract.

hacheigriega:
## Explanation of Changes

This PR makes use of the core contract's distribution message types to report payout allocations in case of data request timeouts. When a data request times out, either due to insufficient number of commits or reveals, we pay some fixed amount to the ones that did commit. This amount is a module parameter called `GasCostCommitment`, whose default value is `100_000_000` (subject to change).

## Testing

Integration testing simulates commit and reveal timeouts under varying replication factors and numbers of commits and reveals. The core contract used for integration testing has been updated to the one built from commit [aa0a52f](https://github.com/sedaprotocol/seda-chain-contracts/commit/aa0a52f339d1bd1ddc30e6d344b4c96de12e5c33). 

## Related PRs and Issues

Closes #453


x/tally/keeper/endblock_test.go
```diff
@@ -10,25 +10,103 @@ import (
...
+			dataResult, err := f.batchingKeeper.GetLatestDataResult(f.Context(), drID)
+			require.NoError(t, err)
+			require.Equal(t, tt.expExitCode, dataResult.ExitCode)
 
-	dataResults, err := f.batchingKeeper.GetDataResults(f.Context(), false)
-	require.NoError(t, err)
-	require.Equal(t, *dataResult, dataResults[0])
+			dataResults, err := f.batchingKeeper.GetDataResults(f.Context(), false)
+			require.NoError(t, err)
+			require.Contains(t, dataResults, *dataResult)
```
hacheigriega, line 107: Was having trouble implementing checks for before & after balances of the executors here. Will look into it soon.

hacheigriega:
I had mentioned that I would add filtering support for requests with missing reveals in this PR, but I think it's better to create a separate PR for that. I'll do that tomorrow.

x/tally/keeper/endblock.go
```diff
@@ -109,31 +121,27 @@ func (k Keeper) ProcessTallies(ctx sdk.Context, coreContract sdk.AccAddress) err
 
 			k.Logger(ctx).Info("completed tally", "request_id", req.ID)
 			k.Logger(ctx).Debug("tally result", "request_id", req.ID, "tally_result", tallyResults[i])
+
+			// TODO
+			distMsgs = types.DistributionMessages{
+				Messages:   []types.DistributionMessage{},
+				RefundType: types.DistributionTypeNoConsensus,
```
gluax, line 128: Maybe we should add one for insufficient commits/reveals :x

x/tally/keeper/endblock_test.go
gluax: HY and I discussed this, just need to alter tests to account for rewards going into pending withdrawl, and the data request poster getting the remaining escrow refunded.

x/tally/keeper/payout.go
```diff
@@ -0,0 +1,60 @@
...
+	}
+	sort.Strings(committers)
+
+	distMsgs := make([]types.DistributionMessage, len(committers))
+	for i, committer := range committers {
+		distMsgs[i] = types.DistributionMessage{
+			Kind: types.DistributionKind{
+				ExecutorReward: &types.DistributionExecutorReward{
+					Identity: committer,
+					Amount:   math.NewIntFromUint64(gasCost),
```
gluax, line 43: Are we doing the calculations we discussed for devnet5 where one executor is paid more, or whatever it was we decided? If so is that coming in another PR?

x/tally/keeper/endblock_test.go
Thomasvdam: Quick 2 cents: I don’t think we should test whether the payout messages get handled correctly on the chain side. The contract unit tests can take care of that and for integration the test suite is probably a better candidate.

gluax: That also works :)

x/tally/keeper/payout.go
Thomasvdam, line 43: We should take the predetermined `gas_cost_commitment` for the commitment and multiply it with the `gas_price` as it was specified in the data request.

And probably we need to separate calculating the gas consumed for all the actors and the actual payout. In theory it's possible to create a DR with a really low `exec_gas_limit` and have stupid executors trying to commit to it (as they could know up front that it's not worth executing). Right now we would end up with a scenario where the distribution messages tally up to more than what was attached to the DR.

x/tally/types/params.go
```diff
@@ -9,6 +9,7 @@ const (
 	DefaultFilterGasCostNone             = 100_000
 	DefaultFilterGasCostMultiplierMode   = 100_000
 	DefaultFilterGasCostMultiplierStddev = 100_000
+	DefaultGasCostCommitment             = 100_000_000
```
Thomasvdam, line 12: This should probably be the same as the DR VM startup cost which is  `5_000_000_000_000` right now, but subject to change as we tweak the gas values :)

Thomasvdam (changes requested):

x/tally/keeper/endblock.go
hacheigriega: Not sure what you mean here? :) The switch statement covers those cases right?

x/tally/types/params.go
hacheigriega: fixed, thanks!!

x/tally/keeper/endblock.go
gluax: Your switch case does! I was thinking about the `RefundType` message. It would be more specific for the end user than saying `No Consensus`.

gluax: Idk what do you think @Thomasvdam?

x/tally/keeper/payout.go
hacheigriega: To complete this PR I multiplied the commit gas cost with the request gas price ([e70a899](https://github.com/sedaprotocol/seda-chain/pull/461/commits/e70a89964bbab3a2ac7c4107cd4e664f17d9d57c))
But @Thomasvdam and I discussed that module should return the gas consumption amounts without multiplying them with the gas price. The contract can then multiply the reported gas amounts with the gas price included in the data request when it processes the payments. The contract would also make sure to pay the data providers and executors in alphabetical order until it runs out of funds. Cc: @gluax

hacheigriega: Currently the contract seems to be missing distribution message types for slashing executors and paying data providers. These should also be supported through distribution messages, right?

gluax: Hmmm, I suppose those are fine to calculate on the contract side since it's a sudo message and no therefore has no gas fee.

However, how is alphabetical order determined for the data providers and executors? Is it by address alphabetical order? How do burn messages fit into that? We'd have to iterate over the list do that on the contract side... once to sort once to do the payouts. I feel like it might be easier to pass the data pre-sorted to the contracts.

I thought we weren't slashing executors, just paying the best actor more? As for paying a data provider you need a new enum on the message type, not a new type of distribution message yea?

x/tally/keeper/endblock.go
hacheigriega: Ah yes I find the current list of distribution types to be confusing.
```
const (
	DistributionTypeTallyReward     DistributionType = "tally_reward"
	DistributionTypeExecutorReward  DistributionType = "executor_reward"
	DistributionTypeTimedOut        DistributionType = "timed_out"
	DistributionTypeNoConsensus     DistributionType = "no_consensus"
	DistributionTypeRemainderRefund DistributionType = "remainder_refund"
)
```
We have to add executor reward can be inferred from the distribution kind. Also, time out and no consensus are outcomes of tally, not distribution types, right? We also have to add distribution kinds for slashing and data provider payout, but I'm not sure what distribution types they would be.

gluax: As for data provider payout I would assume it would be a `DistributionSend` message, while I thought we weren't doing any slashing to executors?

x/tally/keeper/payout.go
hacheigriega: Ah sorry, slashing is not a priority at the moment. You're right, we would need a new enum (or `DistributionKind` right?) for paying a data provider. I think the order can be arbitrary, but I think paying data providers -> paying executors -> burn tally gas would be good? 

As for sorting, you're correct actually. The module will return the distribution messages sorted by their addresses since it works with sorted lists of reveals and proxy keys.

gluax: Ok yea that makes sense to me.

Thomasvdam: Probably good to get more inputs on the ordering. I think there’s an argument to be made to have the burn as the highest priority, as this effectively raises the value of the SEDA that the other actors are already holding or will earn in the future. :p

hacheigriega:
I believe this PR is ready to be merged? I will create an issue on the contract repo about what we discussed. Cc @Thomasvdam @gluax

Thomasvdam (approved):

gluax (approved):

PR #462, opened 2025-01-13
Tally requests with missing reveals

hacheigriega:
We now tally requests that have timed out due to missing reveals. Median or mode is calculated the same way as before with whatever reveals that have been collected. However, the threshold for consensus is now 2/3 of the replication factor, instead of the number of collected reveals. 

In addition, the filter functions have been fully separated from the tally keeper type for modularity. The filter and tally process has also been slightly refactored for clarity.


Thomasvdam (approved):
lgtm :)

Issue #459, opened 2025-01-07
✨ Batch fraud proof


hacheigriega:
## ✨ Feature

In the batching module, create a transaction endpoint for submitting a batch fraud proof. Then, create a begin blocker for handling a fraud proof, much like how the SDK's evidence module handles an equivocation evidence.

PR #463, opened 2025-01-13
Add double batch signing evidence submission

Thomasvdam:
## Motivation

Allow actors to submit a transaction with evidence of a validator signing 2 different batches from the same height. We treat this the same way as double block signing (equivocation):
- The validator gets slashed for the same amount as an equivocation
- The validator gets tombstoned
- The validator gets jailed

## Explanation of Changes

Original plan as discussed:

1. Submit (batchHeight, operatorAddress, resultRoot, validatorRoot, metadataRoot, signature, provingScheme)
2. Retrieve batch by batchHeight, validate it exists
3. Compute fraudulent batchID from the submitted batch height and roots.
4. Validate fraudulent batchID is different from batch.BatchID
5. Retrieve signature for batch by operatorAddress, validate it exists
6. Validate signatures for validator are different
7. Retrieve previous batch by batchHeight - 1
8. Retrieve public key for validator by operatorAddress from previous batch
9. Perform signature verification with submitted signature and retrieved public key, verify it’s valid
10. Retrieve validator set from height of batch (stakingKeeper historical info)
11. Get power of validator at the time of the double signed batch (required to determine slash amount)
12. Do basically the same thing as the Equivocation handler https://github.com/cosmos/cosmos-sdk/blob/v0.50.9/x/evidence/keeper/infraction.go

I skipped 6 since the extra check doesn't give us any information we don't also get by simply comparing the recovered public key (eth address) to the known one.

I do the slashing through the stakingkeeper as the slashing keeper only allows us to mark the slash as `unspecified | double_block | liveness`.

## Testing

Added integration tests for the keeper and unit tests for the evidence validation and batchID generation. There are a couple of paths that are not tested since they don't really make sense. Mainly certain keepers returning errors or unexpected results, or the address derivation from a signature while we just got the address from a library that should already have errored on invalid input.

I also tested it against a local chain by taking the generated SEDA keys file and signing a manually constructed batch for height 2. Before taking the keys I rotated the key file so there was a second batch.

## Related PRs and Issues

Closes: #459


x/batching/keeper/evidence.go
```diff
@@ -0,0 +1,217 @@
...
+	sdkCtx.Logger().Info("confirmed double batch sign", "validator", evidence.OperatorAddress, "batch height", evidence.BatchHeight, "block height", evidence.BlockHeight)
+
+	// Reject evidence if the double-sign is too old. Evidence is considered stale
+	// if the difference in number of blocks is greater than the allowed
+	// parameter defined.
+	infractionHeight := evidence.GetHeight()
+	ageBlocks := sdkCtx.BlockHeader().Height - infractionHeight
+	cp := sdkCtx.ConsensusParams()
+	if cp.Evidence != nil {
+		if ageBlocks > cp.Evidence.MaxAgeNumBlocks {
```
hacheigriega, line 79: Technically this will never happen because the historical info look up will fail after 10,000 blocks (`historical_entries` parameter) while this `max_age_num_blocks` param is 100,000, right?

x/batching/keeper/evidence.go
```diff
@@ -0,0 +1,217 @@
...
+	sdkCtx.Logger().Info("received batch double sign evidence", "batch height", evidence.BatchHeight, "operator address", evidence.OperatorAddress, "result root", evidence.DataResultRoot, "validator root", evidence.ValidatorRoot, "proving metadata hash", evidence.ProvingMetadataHash, "proving scheme index", evidence.ProvingSchemeIndex)
+
+	// Validate that a batch exists for the given batch height.
+	batch, err := k.GetBatchByBatchNumber(ctx, evidence.BatchHeight)
+	if err != nil {
+		return err
+	}
+
+	if batch.BlockHeight != evidence.BlockHeight {
+		return fmt.Errorf("batch height mismatch. Expected: %d, Got: %d", batch.BlockHeight, evidence.BlockHeight)
```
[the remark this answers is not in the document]
Thomasvdam: Yes, I added it to provided a little more feedback on which part is wrong rather than just having it fail on an invalid signature. No strong attachment though, open to removing it :)

x/batching/keeper/evidence.go
Thomasvdam: Hmm, that's a pity. The block equivocation handler relies on the power as it is submitted on the evidence, but we have no way of really verifying the power if we let a user submit it in the TX.

I suppose we can drop this check and replace it with the historical entries from the staking keeper?

x/batching/keeper/evidence.go
```diff
@@ -0,0 +1,216 @@
...
+		return fmt.Errorf("recovered address does not match validator entry. Recorded: %s, Got: %s", hex.EncodeToString(validatorEthAddr), hex.EncodeToString(signatureAddr))
+	}
+
+	sdkCtx.Logger().Info("confirmed double batch sign", "validator", evidence.OperatorAddress, "batch number", evidence.BatchNumber, "block height", evidence.BlockHeight)
+
+	// Reject evidence if the double-sign is too old. Evidence is considered stale
+	// if the difference in number of blocks is greater than the allowed
+	// parameter defined.
+	// NOTE: The default max age is way larger than the default historical entries
+	// we should check how CometBFT valdiates the power attached to block level evidence
```
mariocao, line 72:
```suggestion
	// we should check how CometBFT validates the power attached to block level evidence
```

x/batching/keeper/evidence.go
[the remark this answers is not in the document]
Thomasvdam: Turns out this was an attack vector so we removed it completely 😅

x/batching/keeper/evidence.go
hacheigriega: Found that CometBFT uses the state store to load the validator set at the time of infraction (see https://github.com/cometbft/cometbft/blob/9722b6dfa1d0da3c3d4f73c1c779479971959d94/evidence/verify.go#L51) but I wasn't able to figure out whether this store is accessible from the application layer. I don't think it is.

hacheigriega (approved):

Issue #458, opened 2025-01-07
✨ Data provider payout


hacheigriega:
## ✨ Feature

- Infer the amount of gas used by the data providers based on their fees stored in the data proxy module and the gas price attached to the data request.
- Subtract this inferred amount from the amount of gas used by the executors (calculated from: #456). Then, construct two payout lists for data providers and executors, respectively.
- Attach the payout lists to the sudo call to the core contract, which would then process the actual payments.
- Burn the fees for the tally VM gas to complete the incentives implementation.

hacheigriega:
Closed by https://github.com/sedaprotocol/seda-chain/pull/464

PR #464, opened 2025-01-13
Executor & data provider payouts

### LINKED ISSUE #456: ✨ Data request executor payout & slashing
## ✨ Feature

- Compute the "adjusted gas usage" as needed and, together with the state returned by the filtering process, compute payouts to each of the executors. The mechanism is described in detail [here](https://www.notion.so/sedaprotocol/Executor-Node-Gas-Reporting-and-Payout-Mechanism-160a68d575ca807e8276d04534a430fc?pvs=4).
- Slash executors that have reported outlier data if consensus has been reached in the filtering process. This process would involve identifying the addresses of the executors that have reported outlier data and making a sudo call to the core contract to slash their stakes.

hacheigriega:
## Motivation

Implementation of data executor payout calculation and total execution gas consumption calculation in both uniform and divergent gas reporting scenarios.

See [this](https://www.notion.so/sedaprotocol/x-tally-161a68d575ca8032ad1fd4714a48ec4b?pvs=4#17ca68d575ca808dbe92ff97cfec6bd5) table for broad descriptions of the payout logic. 

## Explanation of Changes

- [x] Executor payout (uniform & divergent cases)
- [x] Data provider payout
- [x] Re-ordering error detection
- [x] Add base fee, which is charged for every data request
- [x] Test with updated core contract
- [x] Burn 20% of executor payout in case of `ErrInvalidFilterInput`, `ErrNoConsensus`, or a tally VM execution error
- [x] Add more tests (particularly ones with proxy payouts)
- [x] Combine burn messages into one and send distributions in a sorted way (burn -> data proxies -> executors)
- [x] Add events to show details of payout calculation outcomes

## Related PRs and Issues

Closes #456 & #458


x/tally/keeper/payout.go
```diff
@@ -35,27 +58,114 @@ func (k Keeper) CalculateCommitterPayouts(ctx sdk.Context, req types.Request, ga
...
+			payout = adjPayout
+		}
+		dists[i] = types.NewExecutorReward(executor, payout)
+
+		ctx.EventManager().EmitEvent(sdk.NewEvent(types.EventTypeExecutorRewardUniform,
+			sdk.NewAttribute(types.AttributeExecutor, executor),
+			sdk.NewAttribute(sdk.AttributeKeyAmount, payout.String()),
+		))
+	}
+	return dists, totalGasUsed, totalBurn
```
hacheigriega, line 148: Note that in the current implementation, the total gas used does not change even in case of reduced payout.

x/tally/keeper/payout.go
```diff
@@ -35,27 +58,114 @@ func (k Keeper) CalculateCommitterPayouts(ctx sdk.Context, req types.Request, ga
...
+		dists[i] = types.NewExecutorReward(executor, payoutAmt)
+
+		ctx.EventManager().EmitEvent(sdk.NewEvent(types.EventTypeExecutorRewardUniform,
+			sdk.NewAttribute(types.AttributeExecutor, executor),
+			sdk.NewAttribute(sdk.AttributeKeyAmount, payoutAmt.String()),
+		))
 	}
-	result.Messages = distMsgs
-	return result, nil
+	return dists, adjGasUsed * uint64(replicationFactor), totalBurn
```
hacheigriega, line 104: ditto

x/tally/keeper/endblock.go
```diff
@@ -124,7 +127,13 @@ func (k Keeper) ProcessTallies(ctx sdk.Context, coreContract sdk.AccAddress) err
 			k.Logger(ctx).Debug("tally result", "request_id", req.ID, "tally_result", tallyResults[i])
 		}
 
-		processedReqs[req.ID] = distMsgs
+		baseFee := gasPrice.Mul(math.NewIntFromUint64(params.GasCostBase))
```
Thomasvdam, line 130: I'll leave comments as I go through the PR rather than submitting them all at once.

I don't think this is correct. We should always start with the base fee so if the requestor did not attach enough gas the tally phase will run out of gas as expected during VM execution.

x/tally/keeper/endblock.go
```diff
@@ -124,7 +127,13 @@ func (k Keeper) ProcessTallies(ctx sdk.Context, coreContract sdk.AccAddress) err
...
+		processedReqs[req.ID] = record.DistributionsWithBaseFee(baseFee)
+
+		ctx.EventManager().EmitEvent(sdk.NewEvent(types.EventTypeBaseFeeBurn,
+			sdk.NewAttribute(sdk.AttributeKeyAmount, baseFee.String()),
+		))
```
Thomasvdam, line 135: It would be easier if all attributes get emitted under a single event. If that's too much of a headache than we should at least add the DR_ID as an attribute so we can link the event to the relevant DR.

x/tally/keeper/endblock.go
```diff
@@ -199,25 +206,45 @@ func (k Keeper) FilterAndTally(ctx sdk.Context, req types.Request, params types.
...
-	result.Consensus = filterResult.Consensus
-	result.ProxyPubKeys = filterResult.ProxyPubKeys
-	result.TallyGasUsed += filterResult.GasUsed
+	// Begin tracking tally result and burn amount.
+	result := TallyResult{
+		Consensus:    filterResult.Consensus,
+		ProxyPubKeys: filterResult.ProxyPubKeys,
+		TallyGasUsed: filterResult.GasUsed,
+	}
+	burnAmt := math.ZeroInt()
```
Thomasvdam, line 217: I think in phase 1 we don't have to consider the amount burned yet and we can simply deal with gas units. Only once we get ready to send it to the contract do we convert to `aseda`.

x/tally/keeper/endblock.go
```diff
@@ -199,25 +206,45 @@ func (k Keeper) FilterAndTally(ctx sdk.Context, req types.Request, params types.
...
 
-	// Phase II: Tally Program Execution
+	// Phase 2: Tally Program Execution
+	var vmRes tallyvm.VmResult
+	var tallyErr error
 	if filterErr == nil {
-		vmRes, err := k.ExecuteTallyProgram(ctx, req, filterResult, reveals)
-		if err != nil {
-			result.Result = []byte(err.Error())
+		vmRes, tallyErr = k.ExecuteTallyProgram(ctx, req, filterResult, reveals)
```
Thomasvdam, line 229: The execution needs to be aware of the base gas fee that was already consumed.

x/tally/keeper/payout.go
```diff
@@ -10,22 +11,44 @@ import (
...
+// CalculateDataProxyPayouts returns payouts for the data proxies and
+// returns the the gas used by the data proxies per executor.
+func (k Keeper) CalculateDataProxyPayouts(ctx sdk.Context, proxyPubKeys []string, gasPrice math.Int) ([]types.Distribution, uint64) {
+	if len(proxyPubKeys) == 0 {
+		return nil, 0
 	}
-	if len(req.Commits) == 0 {
-		return result, nil
+	gasUsed := math.ZeroInt()
+	dists := make([]types.Distribution, len(proxyPubKeys))
```
Thomasvdam, line 21: Should we not pass the capacity as not all 'slots' of the slice are guaranteed to be filled? And change the implementation to use append? Or is it ok to have the calling code be responsible for dealing with empty 'slots'?

x/tally/keeper/payout.go
```diff
@@ -10,22 +11,44 @@ import (
...
+			continue
+		}
+
+		gasUsed = gasUsed.Add(proxyConfig.Fee.Amount.Quo(gasPrice))
+		dists[i] = types.NewDataProxyReward(pubKeyBytes, proxyConfig.Fee.Amount)
+
+		ctx.EventManager().EmitEvent(sdk.NewEvent(types.EventTypeDataProxyReward,
+			sdk.NewAttribute(types.AttributeProxyPubKey, pubKey),
+			sdk.NewAttribute(sdk.AttributeKeyAmount, proxyConfig.Fee.Amount.String()),
+		))
```
Thomasvdam, line 40: See other comment about events 😬

x/tally/keeper/payout.go
```diff
@@ -35,27 +58,110 @@ func (k Keeper) CalculateCommitterPayouts(ctx sdk.Context, req types.Request, ga
...
-func calculateExecGasUsed(reveals []types.RevealBody) uint64 {
-	var execGasUsed uint64
-	for _, reveal := range reveals {
-		execGasUsed += reveal.GasUsed
+func median(arr []uint64) uint64 {
+	sort.Slice(arr, func(i, j int) bool {
+		return arr[i] < arr[j]
+	})
+	return arr[len(arr)/2]
+}
```
Thomasvdam, line 152: Should we not handle even numbered lists by taking the average of the middle 2 elements? Or are we ok with one of the 2 values rather than the mean of those? If so we should probably not call the method median to prevent confusion.

proto/sedachain/tally/v1/tally.proto
```diff
@@ -15,7 +15,8 @@ message Params {
   // FilterGasCostMultiplierStdDev is the gas cost multiplier for a filter type
   // stddev.
   uint64 filter_gas_cost_multiplier_std_dev = 4;
-  // GasCostCommitment is the gas cost for a commitment corresponding to an
-  // expired data request.
-  uint64 gas_cost_commitment = 5;
+  // GasCostBase is the base gas cost for a data request.
+  uint64 gas_cost_base = 5;
+  // GasCostCommit is the gas cost for a commit charged under certain scenarios.
+  uint64 gas_cost_commit = 6;
```
mariocao, line 21: where is this used?

x/tally/keeper/payout.go
```diff
@@ -35,27 +58,110 @@ func (k Keeper) CalculateCommitterPayouts(ctx sdk.Context, req types.Request, ga
...
+
+// CalculateUniformPayouts calculates payouts for the executors when their gas
+// reports are uniformly at "gasReport". It also returns the total execution gas
+// consumption and, in case of reduced payout, the amount to be burned.
+func (k Keeper) CalculateUniformPayouts(ctx sdk.Context, executors []string, gasReport, execGasLimit uint64, replicationFactor uint16, gasPrice math.Int, reducedPayout bool) ([]types.Distribution, uint64, math.Int) {
+	adjGasUsed := min(gasReport, execGasLimit/uint64(replicationFactor))
+	payoutAmt := gasPrice.Mul(math.NewIntFromUint64(adjGasUsed))
+	burnAmt := math.ZeroInt()
+	if reducedPayout {
+		burnRatio := math.LegacyNewDecWithPrec(2, 1) // burn 20%
```
Thomasvdam, line 82: This should probably be a module parameter.

proto/sedachain/tally/v1/tally.proto
hacheigriega: We charge the base gas cost for every data request.

When the request times out or when there is no basic consensus, we charge the commit cost for each executor that did commit (`CalculateCommitterPayouts()`)

x/tally/keeper/payout.go
```diff
@@ -35,27 +58,110 @@ func (k Keeper) CalculateCommitterPayouts(ctx sdk.Context, req types.Request, ga
...
+			burnAmt := burnRatio.MulInt(payout).TruncateInt()
+			totalBurn = totalBurn.Add(burnAmt)
+			payout = payout.Sub(burnAmt)
 		}
+		dists[i] = types.NewExecutorReward(executor, payout)
+
+		ctx.EventManager().EmitEvent(sdk.NewEvent(types.EventTypeExecutorRewardUniform,
+			sdk.NewAttribute(types.AttributeExecutor, executor),
+			sdk.NewAttribute(sdk.AttributeKeyAmount, payout.String()),
+		))
```
Thomasvdam, line 142: In terms of events I think we're only really interested in 2 values: `medianGas` and `lowestReporter` (which can be empty for uniform scenarios). This will allow us to simple add those attributes to the DR completion event and simplify indexing in that regard. The adjusted values after the shares calculation we will receive from the contract when it sends the actual SEDA amounts to the identities.

x/tally/keeper/endblock.go
```diff
@@ -227,14 +254,48 @@ func (k Keeper) FilterAndTally(ctx sdk.Context, req types.Request, params types.
...
+		for i, reveal := range reveals {
+			if proxyGasUsedPerExec > reveal.GasUsed {
+				gasReports[i] = 0
+			} else {
+				gasReports[i] = reveal.GasUsed - proxyGasUsedPerExec
+			}
+		}
+
+		var execGasUsed uint64
+		if req.ReplicationFactor == 1 || areGasReportsUniform(gasReports) {
```
Thomasvdam, line 286: Can we not add the RF=1 case to `areGasReportsUniform`? Since we check for 0 there as well.

x/tally/keeper/payout.go
```diff
@@ -35,27 +58,110 @@ func (k Keeper) CalculateCommitterPayouts(ctx sdk.Context, req types.Request, ga
...
+		if i == 0 || adjGasUsed[i] < lowestGasUsed {
+			lowestReporterIndex = i
+			lowestGasUsed = adjGasUsed[i]
+		}
+	}
+	medianGasUsed := median(adjGasUsed)
+	totalGasUsed := medianGasUsed*uint64(replicationFactor-1) + min(lowestGasUsed*2, medianGasUsed)
+	totalShares := medianGasUsed*uint64(replicationFactor-1) + lowestGasUsed*2
+	lowestPayout := gasPrice.Mul(math.NewIntFromUint64(lowestGasUsed * 2 * totalGasUsed / totalShares))
+	normalPayout := gasPrice.Mul(math.NewIntFromUint64(medianGasUsed * totalGasUsed / totalShares))
```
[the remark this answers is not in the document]
hacheigriega: Are you referring to the formula that calculates `totalGasUsed`? That one matches the spec.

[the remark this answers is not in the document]
Thomasvdam: You're right, I missed the `min()` on 118 :)

x/tally/keeper/endblock.go
hacheigriega: Shouldn't the base gas be taken from the execution gas limit while the tally program uses the tally gas limit?

hacheigriega: OK added. Perhaps `BuildFilter()` should check if the filter gas cost is within the tally limit?

x/tally/keeper/payout.go
hacheigriega: Great catch, fixed.

x/tally/keeper/payout.go
hacheigriega: ah right, fixed to match the definition of median (int division for middle 2 elements)

x/tally/keeper/endblock.go
hacheigriega: Changed the check in `areGasReportsUniform` to 
```
if len(reports) <= 1 {
		return true
}
```

x/tally/keeper/endblock.go
hacheigriega: `DistributionsFromGasCalculation()` constructs a series of event attributes and emits them at once at the end. Since an attribute is just a key-value pair, proxy & executor gas attributes' values follow the "ID,amount" pattern.

x/tally/keeper/payout.go
hacheigriega: Please take a look at the function `DistributionsFromGasCalculation()` and let me know what you think. If we want those values, we can add those values to the `GasCalculation` struct and emit them as events later, too.

proto/sedachain/tally/v1/tally.proto
Thomasvdam: Having given it some though maybe it's good to have the name be a little more descriptive of when it's used? I feel like that current name implies that it's a base fee we're charging for the commitment, but in reality it's the fallback gas consumption when we have no other way of determining how much gas was used.

x/tally/keeper/endblock.go
```diff
@@ -103,20 +103,26 @@ func (k Keeper) ProcessTallies(ctx sdk.Context, coreContract sdk.AccAddress) err
...
+			gasCalc.Burn += params.GasCostBase
 		case len(req.Reveals) == 0:
 			dataResults[i].Result = []byte("no reveals")
 			dataResults[i].ExitCode = TallyExitCodeNoReveals
 			k.Logger(ctx).Info("data request has no reveals", "request_id", req.ID)
-			record.ExecDists = k.CalculateCommitterPayouts(ctx, req, params.GasCostCommit, gasPrice)
+
+			gasCalc.Executors, dataResults[i].GasUsed = CommitGasUsed(req, params.GasCostCommit, req.ExecGasLimit)
+			dataResults[i].GasUsed += params.GasCostBase
+			gasCalc.Burn += params.GasCostBase
```
Thomasvdam, line 123: I'm not sure if this case makes sense? If there's no reveals, but we know that there are enough commits to meet the replication factor then there's either a replication factor of 0 (which the contract doesn't allow if I'm not mistaken), or all executors are refusing to reveal which should result in slash events for all of them.

x/tally/types/params.go
```diff
@@ -13,6 +17,8 @@ const (
 	DefaultGasCostCommit                 = 5_000_000_000_000
 )
 
+var DefaultBurnRatio = math.LegacyNewDecWithPrec(2, 1)
```
Thomasvdam, line 20: Is there a reason we have this as a var and the rest as a const?

x/tally/keeper/endblock.go
```diff
@@ -103,20 +103,26 @@ func (k Keeper) ProcessTallies(ctx sdk.Context, coreContract sdk.AccAddress) err
 			return fmt.Errorf("invalid gas price: %s", req.GasPrice)
 		}
 
-		var record types.PayoutRecord
+		var gasCalc GasCalculation
```
Thomasvdam, line 106: This may be more of a personal preference thing, but I feel like keeping track of gas makes more sense as a meter rather than a tally. Similar to prepaid phone credits and a subscription plan. With a plan they'll let your call minutes count up as they already have your bank details and can simply charge the incurred cost to your account. With prepaid they'll count down because they want to cut the call when your credit runs out.

Practically speaking starting at a limit and counting down to 0 is the same as starting at 0 and counting up towards a limit, counting down feels less error prone. For example, when we need to check if a certain action fits in the remaining gas it's easier when you count down: `cost_of_thing <= remaining_limit` v.s. `cost_of_thing <= limit-used_gas`.

Regardless of which direction we go, we probably want to add some sort struct around it so we can easily check if we should proceed. `meter.ConsumeGas(amount)` and if it errors/returns false it means there was not enough gas to execute the operation and we can abort. This means less manual bookkeeping and checks. RIght now for example we first execute the filter before we check if there was even enough gas available to cover the base fee.

x/tally/keeper/gas_calculation.go
```diff
@@ -0,0 +1,222 @@
...
+	PublicKey string
+	Amount    math.Int
+}
+
+// DistributionsFromGasCalculation constructs a list of distribution messages
+// to be sent to the core contract based on the given gas calculation.
+func (k Keeper) DistributionsFromGasCalculation(ctx sdk.Context, reqID string, gasCalc GasCalculation, gasPrice math.Int, burnRatio math.LegacyDec) []types.Distribution {
+	dists := []types.Distribution{}
+	attrs := []sdk.Attribute{
+		sdk.NewAttribute(types.AttributeDataRequestID, reqID),
```
Thomasvdam, line 39: I also need the blockheight at which the DR was issued to be able to uniquely identify which DR this gas result belongs to 😬

x/tally/keeper/gas_calculation.go
```diff
@@ -0,0 +1,222 @@
...
+	for _, proxy := range gasCalc.Proxies {
+		proxyDist := types.NewDataProxyReward(proxy.PublicKey, proxy.Amount, gasPrice)
+		dists = append(dists, proxyDist)
+		attrs = append(attrs, sdk.NewAttribute(types.AttributeDataProxyGas,
+			fmt.Sprintf("%s,%s", hex.EncodeToString(proxy.PublicKey), proxy.Amount.String())))
+	}
+
+	// Append distribution messages for executors, burning a portion of their
+	// payouts in case of a reduced payout scenario.
+	addBurn := math.ZeroInt()
```
Thomasvdam, line 58: Maybe rename this `reducedPayoutBurn` or something along those lines?

x/tally/keeper/gas_calculation.go
```diff
@@ -0,0 +1,222 @@
...
+		executorDist := types.NewExecutorReward(executor.PublicKey, payoutAmt, gasPrice)
+		dists = append(dists, executorDist)
+		attrs = append(attrs, sdk.NewAttribute(types.AttributeExecutorGas,
+			fmt.Sprintf("%s,%s", executor.PublicKey, payoutAmt.String())))
+	}
+
+	// Add additional burn.
+	dists[0].Burn.Amount = dists[0].Burn.Amount.Add(addBurn.Mul(gasPrice))
+
+	ctx.EventManager().EmitEvent(sdk.NewEvent(types.EventTypeGasCalculation, attrs...))
```
Thomasvdam, line 76: If possible I'd like to have the additional burn amount as a separate attribute in the events.

x/tally/keeper/payout.go
Thomasvdam: I think I can determine those values when indexing the event, so if it means we don't have to make big changes we can leave the implementation as is on the chain side.

x/tally/keeper/endblock.go
Thomasvdam: I sort of addressed this in the wall of text about gas metering. TL;DR: we're missing a few gas checks before we start an operation.

x/tally/keeper/gas_calculation.go
hacheigriega: added a new event attribute "reduced_payout_burn"

x/tally/keeper/endblock.go
hacheigriega: We should probably remove this case. However that won't change the executor payout because we pay commit gas to the committers when basic consensus is not reached.

hacheigriega: Perhaps we should change the logic of `CommitGasUsed` so that we only pay the commit gas to the executors that have both committed AND revealed? If someone has committed but failed to reveal, we don't pay the person anything.

Thomasvdam: Yes, except for the case where there are no reveals at all (and the replication factor wasn't met). In that scenario we still attribute them the fallback gas consumption.

proto/sedachain/tally/v1/tally.proto
hacheigriega: agreed and changed the name to `GasCostFallback`

x/tally/keeper/endblock.go
hacheigriega: Thanks, I think the code quality has improved with this change [here]([1095d9f](https://github.com/sedaprotocol/seda-chain/pull/464/commits/1095d9f9a6e08b354e8c26086411fe8a20bbd9eb))

x/tally/keeper/endblock.go
hacheigriega: Addressed this issue in https://github.com/sedaprotocol/seda-chain/pull/464/commits/10fd40b64d57075372ee6fdcba435fef165f046e Please let me know if you still see some problems

x/tally/keeper/gas_meter.go
```diff
@@ -0,0 +1,190 @@
...
+		gasUsed := gasUsedPerExec * uint64(replicationFactor)
+
+		gasMeter.ConsumeExecGasForProxy(pubKeyBytes, gasUsed)
+	}
+}
+
+// MeterExecutorGasFallback computes and records the gas consumption of committers
+// of a data request when basic consensus has not been reached. If checkReveal is
+// set to true, it will only consume gas for committers that have also revealed.
+func MeterExecutorGasFallback(req types.Request, gasCostCommit uint64, gasMeter *types.GasMeter, checkReveal bool) {
```
Thomasvdam, line 95: I think we can infer checkReveal. I can't think of a reason where we would want to reward a fallback to an executor who could have revealed but didn't.

x/tally/keeper/gas_meter.go
```diff
@@ -0,0 +1,190 @@
...
+func MeterExecutorGasUniform(executors []string, gasReport uint64, replicationFactor uint16, gasMeter *types.GasMeter) {
+	gasUsed := min(gasReport, gasMeter.RemainingExecGas()/uint64(replicationFactor))
+	for _, executor := range executors {
+		gasMeter.ConsumeExecGasForExecutor(executor, gasUsed)
+	}
+}
+
+// MeterExecutorGasDivergent computes and records the gas consumption of executors
+// when their gas reports are divergent.
+func MeterExecutorGasDivergent(executors []string, gasReports []uint64, replicationFactor uint16, gasMeter *types.GasMeter) {
```
Thomasvdam, line 140: I think these 2 functions should consider the total proxy gas consumed when calculating the executor gas used.

Let assume a uniform reporting scenario:
```
replication_factor: 10
exec_gas_limit: 1000
reported_gas: 9
data_proxy_gas_cost: 5 -> data_proxy_gas_consumed: 50
```

The current implementation would give us the following executor gas used:
```
gasUsed := min(gasReport, (exec_gas_limit - data_proxy_gas_consumed) / replication_factor))
gasUsed := min(9, (1000 - 50) / 10)
gasUsed := min (9, 95)
gasUsed := 9
```

While it should be:
```
gasUsed := min(gasReport - data_proxy_gas_cost, (exec_gas_limit - data_proxy_gas_consumed) / replication_factor))
gasUsed := min(9 - 5, (1000 - 50) / 10))
gasUsed := min(4, 95)
gasUsed := 4
```

x/tally/keeper/gas_meter.go
```diff
@@ -0,0 +1,190 @@
...
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+
+	"github.com/sedaprotocol/seda-chain/x/tally/types"
+)
+
+// DistributionsFromGasMeter constructs a list of distribution messages to be
+// sent to the core contract based on the given gas meter. It takes the ID and
+// the height of the request for event emission.
+func (k Keeper) DistributionsFromGasMeter(ctx sdk.Context, reqID string, reqHeight uint64, gasMeter *types.GasMeter, burnRatio math.LegacyDec) []types.Distribution {
```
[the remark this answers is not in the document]
Thomasvdam: Aaah, we can't implement a new method here of course since the type is from a different package. That makes this make more sense. :)

x/tally/keeper/gas_meter.go
Thomasvdam: I see where I went wrong, we do the adjustment elsewhere. I still changed it since right now the implementation relies on no other operations having changed the exec gas which feels a little brittle. Obviously my change also relies on the data proxy calculation to have been done first, but it won't break if we introduce some other tweak to the gas calculation.

Thomasvdam (approved):
Poking fun at my tiny contribution at the end 😬 

lgtm

x/tally/types/gas_meter.go
```diff
@@ -37,8 +37,15 @@ func NewGasMeter(tallyGasLimit, execGasLimit, maxTallyGasLimit uint64, gasPrice
...
 	}
 
-	// Consume base gas cost, ignoring even if gas runs out.
-	gasMeter.ConsumeTallyGas(baseGasCost)
+	// For normal operations we first check if the gas limit is enough to cover
+	// the operation. For the base gas cost we always want to consume it as even
+	// running out of gas incurs a cost to the network.
+	if baseGasCost > gasMeter.tallyGasLimit {
+		gasMeter.tallyGasRemaining = 0
+	} else {
```
hacheigriega, line 45: Isn't this case taken care of by `ConsumeTallyGas()` logic anyways?

Thomasvdam: Good catch, but the problem was that the gas should not be consumed in `ConsumeTallyGas()` as an operation that doesn't 'fit' in the remaining gas limit should not be executed.

PR #474, opened 2025-01-24
Overflow protection and fuzz tests for gas meter & std dev filter

hacheigriega:
- Added overflow protections in tally and pubkey modules.
- Fuzz testing of tally gas meter and standard deviation particularly involving large numbers
- Fix the number parsing of standard deviation filter to follow the pattern of ojg JSON parser, which parses numbers in the range of `-9223372036854775799` and `9223372036854775799` as int64 and the numbers out of that range as strings (encoding/json.Number).


x/tally/keeper/gas_meter_test.go
```diff
@@ -0,0 +1,214 @@
...
+
+		tallySum := math.NewIntFromUint64(gasMeter.TallyGasUsed())
+		tallySum = tallySum.Add(math.NewIntFromUint64(gasMeter.RemainingTallyGas()))
+		require.Equal(t, tallySum.String(), strconv.FormatUint(tallyGasLimit, 10))
+	})
+}
+
+// expectedExecutorGas returns the expected executor gas consumption and the
+// total executor gas used given the reveals.
+func expectedExecutorGas(reveals map[string]types.RevealBody, proxyGasUsed, execGasLimit, rf uint64) (map[string]uint64, uint64) {
```
Thomasvdam, line 123: I think we touched on this briefly, but I don't see a lot of value in testing the output against basically the same implementation 😅 I think the fuzzer should mainly tell us if certain inputs cause a panic, or if the end state after a bunch of fuzzed actions is where we expect it to be.

Thomasvdam (approved):

x/tally/keeper/gas_meter_test.go
hacheigriega: Yeah I agree. I’ll remove individual gas checks before merging

PR #476, opened 2025-01-25
Standard deviation filter refactor

### LINKED ISSUE #306: ✨ Big numbers support for standard deviation filter
## ✨ Feature

The following are the four number types and their corresponding number type bytecode supported by the standard deviation filter:

- `int32` (`0x00`)
- `int64` (`0x01`)
- `uint32` (`0x02`)
- `uint64` (`0x03`): Not fully supported because the ojg package converts large numbers to `json.Number`. which we do not deal with in the standard filter yet. 

First, we need to add a full support for `uint64` by parsing `json.Number`. We can make use of the generic type as follows:
```
		num, ok := data.(int64)
		if !ok {
			// Big numbers are stored as json.Number.
			bigNum, ok := data.(json.Number)
			if !ok {
				corruptQueue = append(corruptQueue, i)
				continue
			}

			var num T
			var err error
			rt := reflect.TypeOf(num)
			switch rt.Kind() {
			case reflect.Int32, reflect.Int64:
				var t int64
				t, err = strconv.ParseInt(bigNum.String(), 10, rt.Bits())
				num = T(t)
			case reflect.Uint32, reflect.Uint64:
				var t uint64
				t, err = strconv.ParseUint(bigNum.String(), 10, rt.Bits())
				num = T(t)
			default:
				return nil, ErrFilterUnexpected // should not be reachable
			}
			if err != nil {
				corruptQueue = append(corruptQueue, i)
				continue
			}
			nums = append(nums, T(num))
		}
```

Second, we need to clarify in the specifications the number ranges we support and possible overflow issues. We should perhaps add an overflow detection wherever it can happen such as when we cast number types or compute the median.

Third, we will add support for `math/big` numbers, which provide arbitrary-precision arithmetics.

### LINKED ISSUE #470: 🔧 Test and double check std-dev filter
## 🔧 Chore

We should double check how we want the std-dev consensus filter to function, and then ensure the std-dev filter functions that way.

## Motivation

During a call/meeting we started to suspect the std-dev filter might not actually be functioning correctly, or there were different interpretations of how it should work.

hacheigriega:
## Explanation of Changes

Implements full standard deviation functionality for the standard deviation filter, with full support for uint256 and int256 ranges. Although the underlying code for the arithmetic is the same, the number type given as a part of the filter input determines the valid range of reveal data:

- `0x00`: 32 bit `[-4294967295, 4294967295]`
- `0x01`: 64 bit `[-18446744073709551615, 18446744073709551615]`
- `0x02`: 128 bit `[-340282366920938463463374607431768211455, 340282366920938463463374607431768211455]`
- `0x03`: 256 bit `[-115792089237316195423570985008687907853269984665640564039457584007913129639935, 115792089237316195423570985008687907853269984665640564039457584007913129639935]`

It parses the reveal data as big integers (arbitrary-precision), computes the sample standard deviation, and designates any reveals that deviate from the mean by more than the sample standard deviation times the given sigma multiplier as outliers.

Note the results can be slightly inaccurate because the mean and the sample standard deviation are truncated integers. To support real number standard deviations, we would need to check whether the big floating point library is deterministic & worth the additional computational burden.

## Testing

Several unit tests

## Related PRs and Issues

Closes #470
Closes #306


x/tally/types/filters.go
```diff
@@ -215,51 +223,61 @@ func detectOutliersSignedInteger[T constraints.Integer](dataList []any, maxSigma
...
+
+	// Fill out the outliers list.
 	var numsInd, nonOutlierCount int
 	for i := range outliers {
 		if len(corruptQueue) > 0 && i == corruptQueue[0] {
 			outliers[i] = true
 			corruptQueue = corruptQueue[1:]
 		} else {
-			if median.IsWithinSigma(nums[numsInd], maxSigma) {
+			if isWithinStdDev(nums[numsInd], mean, stdDev) {
```
mariocao, line 280: We should use sigma in here.

Instead of `stdDev` we could use something like this:

```
maxDev := new(big.Int).Mul(stdDev, big.NewInt(int64(maxSigma)))
```

Then it should be like:
```suggestion
			if isWithinStdDev(nums[numsInd], mean, maxDev) {
```

Thomasvdam: Should we not first convert the max sigma to a number with 6 decimals?

hacheigriega: fixed to 
```go
maxDev := new(big.Int).Mul(stdDev, new(big.Int).SetUint64(maxSigma.WholeNumber()))
```

hacheigriega: I suppose it will be good to support fractional max sigma here? I think we can do some big rational math in `isWithinStdDev`

proto/sedachain/tally/v1/tally.proto
```diff
@@ -34,4 +34,6 @@ message Params {
   ];
   // MaxResultSize is the maximum size of the result of a data request in bytes.
   uint32 max_result_size = 8;
+  // MaxRevealDataLength is the maximum length of a reveal data.
+  uint32 max_reveal_data_length = 9;
```
Thomasvdam, line 38: I don't think we need this parameter. For now at least we probably only want to limit the size of the node we're extracting in the std-dev filter to prevent expensive big number computation. We can leave this as a hardcoded variable, since changing it will most likely require other changes to prevent abuse. What do you think?

x/tally/types/filters.go
```diff
@@ -126,13 +129,15 @@ func NewFilterStdDev(input []byte, gasCostMultiplier uint64, replicationFactor u
...
 	case 0x01: // Int64
-		filter.filterFunc = detectOutliersSignedInteger[int64]
+		filter.filterFunc = detectOutliersInt[int64]
 	case 0x02: // Uint32
-		filter.filterFunc = detectOutliersUnsignedInt[uint32]
+		filter.filterFunc = detectOutliersInt[uint32]
 	case 0x03: // Uint64
-		filter.filterFunc = detectOutliersUnsignedInt[uint64]
+		filter.filterFunc = detectOutliersInt[uint64]
+	case 0x04: // BigInt
```
Thomasvdam, line 139: For completeness I think we should add the `(u)int128` and `(u)int256` here as well, even though it all ends up in the same bucket. Makes it easier if at some point we do distinguish between the types.

proto/sedachain/tally/v1/tally.proto
hacheigriega: I'm down to get rid of reveal data length check altogether since we already have some limit on the reveal size in the contract. And the std dev filter limits the big int size at 256 too.

x/tally/types/filters.go
```diff
@@ -164,33 +170,35 @@ func NewFilterStdDev(input []byte, gasCostMultiplier uint64, replicationFactor u
 // max sigma. If less than 2/3 of the reveals are non-outliers, "no
 // consensus" error is returned as well.
 func (f FilterStdDev) ApplyFilter(reveals []RevealBody, errors []bool) ([]bool, bool) {
-	dataList, _ := parseReveals(reveals, f.dataPath, errors)
+	dataList, _ := parseReveals(reveals, f.dataPath, errors, f.maxDataLen)
 	return f.filterFunc(dataList, f.maxSigma, errors, f.replicationFactor)
 }
 
-func detectOutliersSignedInteger[T constraints.Integer](dataList []any, maxSigma Sigma, errors []bool, replicationFactor uint16) ([]bool, bool) {
```
Thomasvdam, line 171: I think this method still uses the old approach and not actually the standard deviation?

x/tally/types/filters.go
hacheigriega: Added four different input types based on bit length of the big Int encoding: 32, 64, 128, and 256.

x/tally/types/filters.go
hacheigriega: I also renamed max sigma to sigma multiplier.

x/tally/types/filters.go
```diff
@@ -118,148 +115,106 @@ func NewFilterStdDev(input []byte, gasCostMultiplier uint64, replicationFactor u
...
-	case 0x03: // Uint64
-		filter.filterFunc = detectOutliersUnsignedInt[uint64]
+	case 0x00: // 32-bit integer
+		filter.bitLenLimit = 32
+	case 0x01: // 64-bit integer
+		filter.bitLenLimit = 64
+	case 0x02: // 128-bit integer
+		filter.bitLenLimit = 128
+	case 0x03: // 256-bit integer
+		filter.bitLenLimit = 256
```
Thomasvdam, line 132: Sorry to be annoying but I think it makes sense to be explicit about signed/unsigned, if only because in smart contract computation you usually have to be explicit about this. 😬

hacheigriega: Now signed and unsigned types are separated and the signed types have 1 less bit length limit compared to the corresponding unsigned types to account for the sign bit. However, this relies on the assumption of ones' complement as opposed to EVM's two's complement.. So for example, the our minimum of int64 is `
-9223372036854775807` rather than Ethereum's `-9223372036854775808`.

Thomasvdam: Hmm, probably not really significant but it does feel weird :P 

I ran some benchmarks, example of the tests:

```golang
package types_test

import (
	"math/big"
	"testing"
)

func BenchmarkUint256CompareEqual(b *testing.B) {
	value := "115792089237316195423570985008687907853269984665640564039457584007913129639935"
	number := new(big.Int)
	number.SetString(value, 10)

	comparisonValue := "115792089237316195423570985008687907853269984665640564039457584007913129639935"
	comparisonNumber := new(big.Int)
	comparisonNumber.SetString(comparisonValue, 10)

	for i := 0; i < b.N; i++ {
		result := number.Cmp(comparisonNumber)
		_ = result
	}
}

func BenchmarkUint256BitsEqual(b *testing.B) {
	value := "115792089237316195423570985008687907853269984665640564039457584007913129639935"
	number := new(big.Int)
	number.SetString(value, 10)

	maxBits := 256

	for i := 0; i < b.N; i++ {
		result := number.BitLen() > maxBits
		_ = result
	}
}
```

```
goos: darwin
goarch: arm64
pkg: github.com/sedaprotocol/seda-chain/x/tally/types
cpu: Apple M2
BenchmarkUint256CompareEqual-8         	243566791	         4.812 ns/op
BenchmarkUint256BitsEqual-8            	1000000000	         0.2973 ns/op
BenchmarkUint256CompareLarger-8        	387422906	         3.004 ns/op
BenchmarkUint256BitsLarger-8           	1000000000	         0.2929 ns/op
BenchmarkUint256CompareSmaller-8       	266968612	         4.526 ns/op
BenchmarkUint256BitsSmaller-8          	1000000000	         0.2919 ns/op
BenchmarkUint256CompareSmallerTiny-8   	444085474	         2.693 ns/op
BenchmarkUint256BitsSmallerTiny-8      	1000000000	         0.2934 ns/op
BenchmarkInt32CompareEqual-8           	332960641	         3.567 ns/op
BenchmarkInt32BitsEqual-8              	1000000000	         1.123 ns/op
BenchmarkInt32CompareLarger-8          	334739314	         3.598 ns/op
BenchmarkInt32BitsLarger-8             	1000000000	         1.120 ns/op
BenchmarkInt32CompareSmaller-8         	367317375	         3.279 ns/op
BenchmarkInt32BitsSmaller-8            	1000000000	         1.163 ns/op
BenchmarkInt32CompareSmallerTiny-8     	366770817	         3.270 ns/op
BenchmarkInt32BitsSmallerTiny-8        	1000000000	         1.164 ns/op
PASS
ok  	github.com/sedaprotocol/seda-chain/x/tally/types	19.490s
```

It is significantly slower if we simply compare to a precomputed big number, but I'm not sure we'll be running this computation at a scale that this really matters. I think correctness is worth having a little less efficient and more verbose code. What's your take? Also curious what @mariocao thinks :)

mariocao: I agree, let's do this simply precomputed comparison for the sake of correctness. I think the cpu cycles used for that are negligible. 😓

hacheigriega: addressed in [1410351](https://github.com/sedaprotocol/seda-chain/pull/476/commits/14103512de7e0455a88d6fc641af0c8b8b1b9234)!

Thomasvdam (approved):

PR #478, opened 2025-01-27
chore(x/tally): update endblock error handling

Thomasvdam:
I'm the least sure about the fallback values. In the end I decided to just write the values we got from the contract to the store, instead of redoing all the encoding checks to figure out what was wrong. 

## Motivation

Main difference is that we now allow the node to panic if we are unable to write to the store. This prevents us from continuing on invalid state should something go wrong there. It should never happen due to inputs as those get validated at mutliple points throughout the code.

We also write an empty result to the store in case there are encoding/decoding errors for fields that we are in control of.

## Explanation of Changes

N.A.

## Testing

Tests still work, no panics yet as this shouldn't really happen anyway :)

## Related PRs and Issues

N.A.

### DIFF
--- a/x/tally/types/data_result.go
+++ b/x/tally/types/data_result.go
@@ -1,0 +1,22 @@
+package types
+
+import (
+	"fmt"
+
+	"cosmossdk.io/math"
+
+	batchingtypes "github.com/sedaprotocol/seda-chain/x/batching/types"
+)
+
+// MarkResultAsFallback marks a DataResult as a fallback result.
+// This is used when the request cannot be processed due to an error on our side, as all
+// encoding/decoding of values is done by either the contract or the chain.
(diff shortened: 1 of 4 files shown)


gluax (approved):

mariocao (approved):
Lgtm

x/tally/types/data_result.go
```diff
@@ -0,0 +1,22 @@
...
+	"fmt"
+
+	"cosmossdk.io/math"
+
+	batchingtypes "github.com/sedaprotocol/seda-chain/x/batching/types"
+)
+
+// MarkResultAsFallback marks a DataResult as a fallback result.
+// This is used when the request cannot be processed due to an error on our side, as all
+// encoding/decoding of values is done by either the contract or the chain.
```
hacheigriega, line 13: I suppose this could also be a method with pointer receiver?

hacheigriega (approved):

PR #484, opened 2025-02-03
feat: chain seda prometheus metrics

gluax:
## Motivation

Metrics for seda specific operations

## Explanation of Changes

Added the following metrics:
- time of a batch end block
- time of tally end block
- # of tallys in a block
- tally gas burned per block
- number of oracle programs currently uploaded(bugged it currently double counts and I need to figure that out)

I registered the metrics in the `module.go` for the effected modules in the `RegisterServices` method. If there's a better place I should register the metrics let me know.

Similarly, I created the metric variables in the locations that they were used. If we have a preference to where they should go let me know.

Lastly, I removed some redundant if statements, only cause my linter was angry about them.

## Testing

You can seem there in the prometheus metrics endpoint when it's enabled.
image

And also in prometheus itself:
image

However, I wasn't sure how to test the tally metrics. Test suite seemed broken for the latest versions of things :x. So I could only test the timing, and program count ones locally.

## Related PRs and Issues

N/A


x/batching/module.go
```diff
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 
 	"github.com/grpc-ecosystem/grpc-gateway/runtime"
+	"github.com/prometheus/client_golang/prometheus"
```
Thomasvdam, line 8: Is there a reason we're using this directly instead of the Cosmos telemetry package? (which wraps this package as far as I can see).

gluax: Which also makes sense to me? There's two different metrics endpoints for them entirely. Telemetry enables a `/metrics` on the API side. While prometheus has it's own separate port and metrics endpoint.

Thomasvdam: Then were are our current metrics coming from? :P I don't see any calls to this package directly in the SDK code

x/wasm-storage/keeper/msg_server.go
```diff
@@ -63,9 +69,9 @@ func (m msgServer) StoreOracleProgram(goCtx context.Context, msg *types.MsgStore
 			sdk.NewAttribute(types.AttributeOracleProgramHash, hashHex),
 		),
 	)
-	if err != nil {
-		return nil, err
-	}
+
+	// Increment the oracle program count.
+	OracleProgramCountMetric.Inc()
```
Thomasvdam, line 74: Maybe it's running once in CheckTx and again in DeliverTx? If it's annoying to debug I would just drop it, this is a metric we can also easily get from the exploder.

gluax: There was an easy fix :). The context has a method to check if it's in `CheckTx` or not.

x/wasm-storage/keeper/msg_server.go
```diff
@@ -25,6 +26,11 @@ func NewMsgServerImpl(keeper Keeper) types.MsgServer {
 
 var _ types.MsgServer = msgServer{}
 
+var OracleProgramCountMetric = prometheus.NewCounter(prometheus.CounterOpts{
+	Name: "seda_oracle_program_count",
+	Help: "The count of stored oracle programs",
+})
+
```
hacheigriega, line 33: Wouldn't it be nice to also track the total size of oracle programs stored in our chain?

x/batching/module.go
gluax: Aha ok we can get it to work via telemetry, but it requires in the node's `app.toml` under the `telemetry` section:

- `prometheus-retention-time` to be greater than 0
- and `metrics-sink = "prometheus` 

And you still need to enable prometheus to get them to prometheus.

gluax:
I got telemetry to match how prometheus metrics are handled... ish.

I removed the program count one and the total bytes of programs on chain one, because telemetry doesn't have them persisting like prometheus did? I think prometheus maybe also has a time to live for metrics but I'm not sure.

TLDR difference is that metrics with prometheus would persist across a node turning off and on. And for telemetry they don't, at least out of the box. But, I think it's best for now that I don't waste any more time on this and those metrics aren't as important anyways :).

x/tally/module.go
```diff
@@ -144,5 +146,13 @@ func (am AppModule) BeginBlock(_ context.Context) error {
 
 // EndBlock returns the end block logic for the tally module.
 func (am AppModule) EndBlock(ctx context.Context) error {
+	start := time.Now()
+	defer func() { telemetry.SetGauge(float32(time.Since(start).Seconds()), "seda_tally_end_block_time") }()
```
hacheigriega, line 150: There is a time measure metric provided by SDK (ex: https://github.com/cosmos/cosmos-sdk/blob/738f288181180d2e77e04b38871c29cbcf97c5dc/x/gov/keeper/abci.go#L24)
I think this should be useful here

gluax: I used that at first but it didn't play as nicely on the Prometheus side. I don't think it was showing up in Prometheus with that approach? But I don't remember at the moment.

x/tally/keeper/endblock.go
```diff
@@ -72,6 +73,7 @@ func (k Keeper) ProcessTallies(ctx sdk.Context, coreContract sdk.AccAddress) err
 		k.Logger(ctx).Error("[HALTS_DR_FLOW] failed to unmarshal tally-ready data requests", "err", err)
 		return nil
 	}
+	telemetry.SetGauge(float32(len(tallyList)), "seda_tally_end_block_number_of_data_requests_to_tally")
```
hacheigriega, line 76: I think a counter would be more appropriate for this? Like: https://github.com/cosmos/cosmos-sdk/blob/8bfcf554275c1efbb42666cc8510d2da139b67fa/x/staking/keeper/msg_server.go#L366

And I think we should increment one by one every time a tally is completed?

gluax: We don't want a counter here. This one is to emit the number of data request tallies processed in the last block.

You could then use Prometheus to get an avg of dr tallies processed over a period of time. Or just have a live view/graph etc.

x/tally/module.go
hacheigriega: ah I see. We can take a look together tmrw too. But I think it'd be nice to be able to get all the metrics including the SDK ones in a uniform way

gluax: Ah ok I figured how to make it work better with prometheus using that approach. I can push that change now!

x/tally/keeper/endblock.go
hacheigriega: Ah understood. thanks

x/tally/keeper/gas_meter.go
```diff
@@ -56,6 +57,7 @@ func (k Keeper) DistributionsFromGasMeter(ctx sdk.Context, reqID string, reqHeig
 	}
 
 	dists[0].Burn.Amount = dists[0].Burn.Amount.Add(reducedPayoutBurn.Mul(gasMeter.GasPrice()))
+	telemetry.SetGauge(float32(dists[0].Burn.Amount.Uint64()), "seda_tally_end_block_gas_burned")
```
Thomasvdam, line 60: Just to check, does this add to the gauge for every invocation or set it to the value passed?

gluax: Ah yea the prometheus version was doing an Add not a Set let me see if telemetry has an equivalent

gluax: Hmmm we could convert it to seda then set it?

gluax: Ok changes made as we discussed and tested it the way we discussed!

x/batching/module.go
```diff
@@ -144,5 +145,7 @@ func (am AppModule) BeginBlock(_ context.Context) error {
 
 // EndBlock returns the end block logic for the batching module.
 func (am AppModule) EndBlock(ctx context.Context) error {
+	start := telemetry.Now()
+	defer telemetry.ModuleMeasureSince(types.ModuleName, start, "seda_batch_end_block_time")
```
hacheigriega, line 149: Let's just follow the standard if it doesn't make a difference?
```go
defer telemetry.ModuleMeasureSince(types.ModuleName, start, telemetry.MetricKeyEndBlocker)
```

x/tally/keeper/endblock.go
```diff
@@ -80,6 +83,7 @@ func (k Keeper) ProcessTallies(ctx sdk.Context, coreContract sdk.AccAddress) err
 	processedReqs := make(map[string][]types.Distribution)
 	tallyResults := make([]TallyResult, len(tallyList))
 	dataResults := make([]batchingtypes.DataResult, len(tallyList))
+	tera_gas_burned := float32(0)
```
hacheigriega, line 86: Some variable naming convention issues in this file

x/tally/module.go
```diff
@@ -144,5 +145,13 @@ func (am AppModule) BeginBlock(_ context.Context) error {
...
 // EndBlock returns the end block logic for the tally module.
 func (am AppModule) EndBlock(ctx context.Context) error {
+	start := telemetry.Now()
+	defer telemetry.ModuleMeasureSince(types.ModuleName, start, "seda_tally_end_block_time")
+
+	// Set this to zero so it always shows up even if there are no tallies
+	telemetry.SetGauge(0.0, "seda_tally_end_block_number_of_data_requests_to_tally")
+	// We reset the amount of gas burned in the end block
+	// since we only add to it during the processing of tallies
+	telemetry.SetGauge(0.0, "seda_tally_end_block_tera_gas_burned")
```
hacheigriega, line 155: Should we create consts for these two?

hacheigriega (changes requested):

x/batching/module.go
gluax: It does. It would label the metric with not a name we want.

x/tally/keeper/endblock.go
gluax: What is the standard then? I'm not familiar with the golang standard

hacheigriega: Ah The golangci comments show up in the files section of the PR

x/batching/module.go
Thomasvdam: Isn’t the module name attached to it though? As far as I can see all modules use the pattern HY linked and we can differentiatie between those right?

x/tally/module.go
Thomasvdam: I think HY meant consts for the labels? That makes sense to me as we’re using them in multiple places

Thomasvdam:
I'm still a little worried about the potential for non-determinism with the float32 so maybe it's an idea to keep these changes in the PR until we've used them in the performance test.

x/batching/module.go
gluax: We could, but imo it's not as clear what the metric is for.

```
end_blocker{module="staking",quantile="0.5"} 0.06427700072526932
end_blocker{module="staking",quantile="0.9"} 0.06427700072526932
end_blocker{module="staking",quantile="0.99"} 0.06427700072526932
end_blocker_sum{module="staking"} 0.06427700072526932
end_blocker_count{module="staking"} 1
```

Those are the metrics in prometheus showing up from telemetry.

VS the pure prometheus metrics:
```
# HELP go_memstats_heap_objects Number of currently allocated objects. Equals to /gc/heap/objects:objects.
# TYPE go_memstats_heap_objects gauge
go_memstats_heap_objects 250384
cometbft_consensus_step_duration_seconds_bucket{chain_id="seda-1-local",step="NewHeight",le="37.27593720314942"} 1
```

which have descriptive names and possibly also help docs. However telemetry does not let us provide help docs.

x/tally/module.go
gluax: Ah ok for the strings I'll add consts

x/tally/module.go
```diff
@@ -144,5 +145,13 @@ func (am AppModule) BeginBlock(_ context.Context) error {
 
 // EndBlock returns the end block logic for the tally module.
 func (am AppModule) EndBlock(ctx context.Context) error {
+	start := telemetry.Now()
+	defer telemetry.ModuleMeasureSince(types.ModuleName, start, telemetry.MetricKeyEndBlocker)
```
gluax, line 149: IMO The default leaves for very unclear metric names. I added my thoughts on why to the earlier conversation.

x/batching/module.go
Thomasvdam: > I don't think following the SDK's lead here is good. Especially considering how often we ourselves complain about the lack of any decent documentation.

Don't get me wrong, I agree that it's not great, but I think that adhering to the (shitty) standards will be more useful than rolling our own.

Realistically speaking operators will just dump SEDA into a monitoring pool of other Cosmos nodes they already have running and call it a day. I think it's better to have that work out of the box and if there's a need we can add our own keys backed up with proper documentation.

> Or at the very least the the ModuleMeasureSince is variadic so we can add more keys for more clarity to it.

Sadly it will join all the keys arguments with an `_` as a separator to get to a single key. :(

PR #486, opened 2025-02-04
Improve batching module testing

hacheigriega:
- Tests for genesis state export and import
- Tests for batch ID computation
- `ValidateGenesis` that checks the validity of batch ID based on tree roots

### DIFF
--- a/x/batching/types/genesis_test.go
+++ b/x/batching/types/genesis_test.go
@@ -1,0 +1,95 @@
+package types_test
+
+import (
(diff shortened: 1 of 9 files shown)


x/batching/types/genesis_test.go
Thomasvdam, line 1: Should we add some tests with invalid json as well?

Thomasvdam (approved):

mariocao (approved):

PR #490, opened 2025-02-06
Fully refund DRs resolved while contract is paused

Thomasvdam:
## Motivation

To prevent data requests being stuck in the contract while it is paused and potentially resolving a lot later when the data is already stale we continue expiring and processing data requests. Any DR resolved while the contract is paused will get an error result and trigger a full refund for the poster of the DR.

## Explanation of Changes

Support the new contract DR list response which also returns the paused status. If the contract is paused all DRs will be resolved with an error and a full refund.

Cleaned up some test WASM binaries that were not being used.

Refactored the integration tests a little to make the paused test easier to set up.

## Testing

Added an integration test with the latest version of the contract.

## Related PRs and Issues

N.A.


gluax (approved):

x/tally/keeper/endblock.go
```diff
@@ -66,15 +61,17 @@ func (k Keeper) ProcessTallies(ctx sdk.Context, coreContract sdk.AccAddress) err
 	}
 	k.Logger(ctx).Info("non-empty tally list - starting tally process")
 
-	var tallyList []types.Request
-	err = json.Unmarshal(queryRes, &tallyList)
+	var contractQueryResponse types.ContractListResponse
+	err = json.Unmarshal(queryRes, &contractQueryResponse)
```
hacheigriega, line 65: I guess the log message in line 72 would have to be updated too

hacheigriega (approved):

x/tally/types/vm_exit_codes.go
```diff
@@ -0,0 +1,10 @@
+package types
+
+const (
+	TallyExitCodeNotEnoughCommits   = 200 // tally VM not executed due to not enough commits
+	TallyExitCodeInvalidRequest     = 201 // tally VM not executed due to invalid request
+	TallyExitCodeContractPaused     = 202 // tally VM not executed due to contract being paused
+	TallyExitCodeInvalidFilterInput = 253 // tally VM not executed due to invalid filter input
+	TallyExitCodeFilterError        = 254 // tally VM not executed due to filter error
+	TallyExitCodeExecError          = 255 // error while executing tally VM
+)
```
hacheigriega, line 10: We can declare these as uint32 if it helps
```
const (
	TallyExitCodeNotEnoughCommits   uint32 = 200 // tally VM not executed due to not enough commits
	TallyExitCodeInvalidRequest     uint32 = 201 // tally VM not executed due to invalid request
	TallyExitCodeContractPaused     uint32 = 202 // tally VM not executed due to contract being paused
	TallyExitCodeInvalidFilterInput uint32 = 253 // tally VM not executed due to invalid filter input
	TallyExitCodeFilterError        uint32 = 254 // tally VM not executed due to filter error
	TallyExitCodeExecError          uint32 = 255 // error while executing tally VM
)
```

PR #492, opened 2025-02-10
READMEs & cleanup

hacheigriega:
- Add READMEs
- Remove CHANGELOG
- Some variable name changes for clarification
- Public key registration invariant now iterates over `ValidatorsByPower` store instead of `Validator` store to avoid checking for a jailed validator's public key.


Thomasvdam (approved):
Small typo in the staking module readme.

PR #501, opened 2025-02-13
Tally parameter for max tallies per block

hacheigriega:


x/tally/types/params.go
```diff
@@ -31,6 +32,7 @@ func DefaultParams() Params {
 		GasCostBase:                   DefaultGasCostBase,
 		ExecutionGasCostFallback:      DefaultExecutionGasCostFallback,
 		BurnRatio:                     DefaultBurnRatio,
+		MaxTalliesPerBlock:            DefaultMaxTalliesPerBlock,
```
Thomasvdam, line 35: Should we add a check to the params validation method?

Thomasvdam (approved):

PR #502, opened 2025-02-13
No executor payout to outliers

hacheigriega:
We now provide an outlier list when we calculate executor payouts so that outliers do not receive execution payouts. Note that this outlier list is `nil` in the following three filter outcomes:

- No basic consensus
- No consensus
- Filter input error

### DIFF
--- a/x/tally/keeper/gas_meter.go
+++ b/x/tally/keeper/gas_meter.go
@@ -129,18 +129,24 @@ func MeterExecutorGasFallback(req types.Request, gasCostFallback uint64, gasMete
 }
 
 // MeterExecutorGasUniform computes and records the gas consumption of executors
-// when their gas reports are uniformly at "gasReport".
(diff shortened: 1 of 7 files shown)


x/tally/keeper/gas_meter.go
```diff
@@ -130,17 +130,20 @@ func MeterExecutorGasFallback(req types.Request, gasCostFallback uint64, gasMete
 
 // MeterExecutorGasUniform computes and records the gas consumption of executors
 // when their gas reports are uniformly at "gasReport".
```
Thomasvdam, line 132: Maybe good to describe the outlier behaviour in the comment here and lines 144-145

hacheigriega: ah yes, fixed!

Thomasvdam (approved):

PR #510, opened 2025-02-20
Basic batch signing testing through e2e test

hacheigriega:
Since simulation testing does not involve the ABCI consensus process, we need to resort to the e2e test to test the complete sequence of ABCI methods we use for batch signing.
This PR adds two unbond transactions to the e2e to trigger creation of batches. The test waits until the number of signatures collected for each batch equals to the number of validators.

### DIFF
--- a/e2e/e2e_test.go
+++ b/e2e/e2e_test.go
@@ -1,6 +1,9 @@
 package e2e
 
-var runWasmStorageTest = true
+var (
+	runWasmStorageTest = true
+	runBatchingTest    = true
+)
 
 func (s *IntegrationTestSuite) TestWasmStorage() {
 	if !runWasmStorageTest {
@@ -8,4 +11,11 @@ func (s *IntegrationTestSuite) TestWasmStorage() {
 	}
 	s.testWasmStorageStoreOracleProgram()
 	s.testInstantiateCoreContract() // involves gov process
+}
+
+func (s *IntegrationTestSuite) TestBatching() {
+	if !runBatchingTest {
+		s.T().Skip()
+	}
+	s.testUnbond() // to trigger batch creation and signing
 }
--- a/e2e/e2e_unbond_test.go
+++ b/e2e/e2e_unbond_test.go
@@ -1,0 +1,95 @@
+package e2e
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	sdktypes "github.com/cosmos/cosmos-sdk/types"
+	stakingtypes "github.com/cosmos/cosmos-sdk/x/staking/types"
+)
+
+func (s *IntegrationTestSuite) testUnbond() {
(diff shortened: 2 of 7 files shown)


e2e/e2e_unbond_test.go
```diff
@@ -0,0 +1,81 @@
...
+			func() bool {
+				batch, err := queryBatch(s.endpoint, 1)
+				s.Require().NoError(err)
+				return len(batch.BatchSignatures) == len(s.chain.validators)
+			},
+			30*time.Second,
+			5*time.Second,
+		)
+
+		s.execUnbond(s.chain, 1, sdk.ValAddress(val2Addr), val2, standardFees.String(), false)
```
Thomasvdam, line 34: I'm a little confused how this can ever work? The E2E chain only has 2 validators right? So when we unbond the 2nd one we should not get any new blocks.

e2e/e2e_unbond_test.go
```diff
@@ -0,0 +1,81 @@
...
+		val2Addr, err := s.chain.validators[1].keyInfo.GetAddress()
+		s.Require().NoError(err)
+		val2 := val2Addr.String()
+
+		s.execUnbond(s.chain, 0, sdk.ValAddress(val1Addr), val1, standardFees.String(), false)
+		s.Require().Eventually(
+			func() bool {
+				batch, err := queryBatch(s.endpoint, 1)
+				s.Require().NoError(err)
+				return len(batch.BatchSignatures) == len(s.chain.validators)
```
Thomasvdam, line 28: I did a run locally to verify my other question about the unbonding (2nd check indeed never gets called) and I noticed that the while the slices are the same length, the signature fields in the structs are empty.

hacheigriega: That's strange because it works on my side. 
Batch 0 has voting percentages `50000000` and `50000000`
Batch 1 has voting percentages  `49999999` and `50000000`
Batch 2 has voting percentages `50000000` and `50000000`

e2e/e2e_unbond_test.go
Thomasvdam: Never mind I didn't see that it only unbonds a tiny amount

e2e/e2e_unbond_test.go
Thomasvdam: It was probably a misconfiguration on my end since when we tried again the checks get called as expected. 😬

e2e/e2e_test.go
```diff
@@ -8,4 +8,5 @@ func (s *IntegrationTestSuite) TestWasmStorage() {
 	}
 	s.testWasmStorageStoreOracleProgram()
 	s.testInstantiateCoreContract() // involves gov process
+	s.testUnbond()                  // to trigger batch creation and signing
```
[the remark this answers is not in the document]
hacheigriega: created a new test `TestBatching` and put it under there

Thomasvdam (approved):

PR #518, opened 2025-03-11
Test vote extension under malicious proposer scenarios

hacheigriega:
Adds test cases involving malicious proposer scenarios to the existing ABCI++ handler testing. 

- Proposer injects empty or arbitrary vote extension (=> vote extension signature verification fails)
- Proposer injects additional vote (=> comparison to last commit data fails)
- Proposer replaces a vote with some arbitrary vote (=> comparison to last commit data fails)

### DIFF
--- a/app/abci/vote_extension_test.go
+++ b/app/abci/vote_extension_test.go
@@ -2,6 +2,7 @@ package abci
 
 import (
 	"bytes"
+	"encoding/base64"
 	"encoding/json"
 	"fmt"
 	"os"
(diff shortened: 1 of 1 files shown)


Thomasvdam (approved):
I think these tests show that a malicious proposer is unable to modify the injected votes as we're using Cosmos SDK builtin functions that verify the serialised commit info against the commit info as it was recorded in CometBFT.

I stepped through the malicious proposer test cases with the step debugger and the checks from the SDK make sense.

PR #522, opened 2025-03-12
Consume remaining tally gas in out of gas case

hacheigriega:
This PR addresses several issues identified during the audit:
- Consume remaining tally gas in an out of gas case.
- Check that the number of transactions in a proposal is not zero to avoid panics in `ProcessProposal` and in `PreBlocker`.
- When we jail validators without the required keys during proving scheme activation, ensure that they have not already been jailed to avoid an error.

### DIFF
--- a/x/pubkey/keeper/endblock.go
+++ b/x/pubkey/keeper/endblock.go
@@ -141,12 +141,14 @@ func (k Keeper) JailValidators(ctx sdk.Context, keyIndex utils.SEDAKeyIndex) err
 			if err != nil {
 				return err
 			}
-			err = k.slashingKeeper.Jail(ctx, consAddr)
-			if err != nil {
-				return err
+			if !val.IsJailed() {
+				err = k.slashingKeeper.Jail(ctx, consAddr)
+				if err != nil {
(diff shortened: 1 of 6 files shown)


x/pubkey/keeper/endblock.go
```diff
@@ -141,12 +141,14 @@ func (k Keeper) JailValidators(ctx sdk.Context, keyIndex utils.SEDAKeyIndex) err
...
 				return err
 			}
-			err = k.slashingKeeper.Jail(ctx, consAddr)
-			if err != nil {
-				return err
+			if !val.IsJailed() {
+				err = k.slashingKeeper.Jail(ctx, consAddr)
+				if err != nil {
+					return err
+				}
```
Thomasvdam, line 148: I think we can add a test case for this right?

hacheigriega: Added an already jailed validator to the test [b9e5d32](https://github.com/sedaprotocol/seda-chain/pull/522/commits/b9e5d32f768f3fb03b3347fa4e75b3bcd8867451)

PR #527, opened 2025-03-19
RefundTxFee method and ante handler for commit/reveal txs

hacheigriega:
## Motivation

Exempting commit and reveal transactions from gas payment has left the chain vulnerable to DoS spamming of commit reveal transactions. We address the issue by requesting a fee payment upfront then refunding the fee payer at the end of the transaction. Since a transaction can involve multiple messages, we use the ante handler to ensure basic properties about commit/reveal transactions and then use the information transmitted by the Core Contract's commit/reveal message handlers to identify the last message in the tx. This way, we ensure that we always process refund once per tx. 

## Explanation of Changes

The ante handler ensures two things:
1. If a transaction contains a commit or a reveal message, then all other messages in the transaction must also be commit or reveal messages.
2. Each commit or reveal message in a transaction is uniquely identified by its `(data_req_ID, public_key, commit_or_reveal_indicator)`.

The Core Contract then sends a `MsgRefundTxFee` at the end of its commit or reveal execution. The wasm-storage module implements a handler for this message to accomplish the following:
1. Ensure the message was sent from the Core Contract.
2. Decode the transaction bytes available in the context to retrieve the last message and the transaction fee information.
3. If the last message in the transaction matches the `(data_req_ID, public_key, commit_or_reveal_indicator)` sent by the contract, refund the fee payer the whole fee collected in the fee collector module account. Note that if the gas limit exceeds the gas consumed multiplied by 3, we do not refund the fee payer to prevent an attack vector where an attacker specifies a gas limit close to the block gas limit to block other transactions.

## Testing

- [x] Manual testing in a local setup
- [x] Tally integration testing has been updated to include the refund flow.
- [x] Wasm storage module unit testing
- [x] Ante handler unit testing

## Related PRs and Issues

- Chain contract https://github.com/sedaprotocol/seda-chain-contracts/pull/280
- Overlay update


x/wasm-storage/keeper/msg_server.go
```diff
@@ -164,3 +167,53 @@ func (m msgServer) UpdateParams(goCtx context.Context, req *types.MsgUpdateParam
...
+	// If the info submitted by the contract matches the last message in the tx,
+	// refund the gas cost to the fee payer.
+	if msgInfo == fmt.Sprintf("%s,%s,%t", req.DrId, req.PublicKey, true) {
+		feeTx, ok := tx.(sdk.FeeTx)
+		if !ok {
+			return nil, errorsmod.Wrap(sdkerrors.ErrInvalidRequest, "transaction is not a fee tx")
+		}
+
+		// Note the fee may be zero in simulation setting.
+		gasPrice := feeTx.GetFee().QuoInt(math.NewIntFromUint64(ctx.GasMeter().Limit()))
```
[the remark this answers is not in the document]
Thomasvdam: It is not :D

x/wasm-storage/keeper/msg_server.go
```diff
@@ -164,3 +167,53 @@ func (m msgServer) UpdateParams(goCtx context.Context, req *types.MsgUpdateParam
...
+		refund := gasPrice.MulInt(math.NewIntFromUint64(ctx.GasMeter().GasConsumed()))
```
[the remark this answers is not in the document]
hacheigriega: This is a much better idea because we can also avoid calculating gas price.
The updated version refunds the fee payer if the gas limit is within the gas consumed times three. If not, there is no refund at all.

x/wasm-storage/keeper/msg_server.go
```diff
@@ -164,3 +167,55 @@ func (m msgServer) UpdateParams(goCtx context.Context, req *types.MsgUpdateParam
...
+	msgInfo, ok := utils.ExtractCommitRevealMsgInfo(coreContract, msgs[len(msgs)-1])
+	if !ok {
+		return nil, errorsmod.Wrap(sdkerrors.ErrInvalidRequest, "last message in tx is not a commit or reveal")
+	}
+
+	// Refund the fee if the info submitted by the contract matches the last msg.
+	if msgInfo == fmt.Sprintf("%s,%s,%t", req.DrId, req.PublicKey, req.IsReveal) {
+		// Refund the fee if the gas limit is set within a reasonable range.
+		// Note the fee may be zero in simulation setting.
+		if !feeTx.GetFee().IsZero() && ctx.GasMeter().Limit() < ctx.GasMeter().GasConsumed()*3 {
```
Thomasvdam, line 209: If the fee is zero we can skip all other checks right?

Thomasvdam (commented):
I guess the vote extension tests comments are a mistake stage?

x/wasm-storage/keeper/msg_server.go
hacheigriega: We now check if the fee is zero in the earlier stage and return without an error if it is zero.

hacheigriega:
> I guess the vote extension tests comments are a mistake stage?

Yes.. And the ante handler contained a bug when dealing with a tx whose first msg is a non-commit/reveal. I'd appreciate a final review on the ante handler 🙏

app/ante.go
```diff
@@ -33,138 +39,56 @@ func NewAnteHandler(options HandlerOptions) (sdk.AnteHandler, error) {
...
-		}
-
-		switch contractMsg := contractMsg.(type) {
-		case CommitDataResult:
-			result, err := d.queryContract(ctx, coreContract, CanExecutorCommitQuery{CanExecutorCommit: contractMsg})
-			if err != nil {
-				return false
+	seen := make(map[string]bool)
+	commitRevealTx := false
+	for i, msg := range tx.GetMsgs() {
```
Thomasvdam, line 70: I think it's more a personal preference thing where I like to prevent nesting conditions, so do with it what you prefer :)

I believe this also achieves the same behaviour:
```golang
seen := make(map[string]bool)
commitRevealTx := false
for i, msg := range tx.GetMsgs() {
	msgInfo, commitRevealMsg := utils.ExtractCommitRevealMsgInfo(coreContract.String(), msg)
	// The first message determines the type of the tx.
	if i == 0 {
		commitRevealTx = commitRevealMsg
	}

	// We don't allow mixing commit/reveal messages with other messages.
	if commitRevealTx != commitRevealMsg {
		return sdk.Context{}, errorsmod.Wrapf(sdkerrors.ErrInvalidRequest, "commit or reveal message cannot be mixed with other messages")
	}

	// If we don't have a commit/reveal message, we can skip the rest of the loop.
	if !commitRevealTx {
		continue
	}

	// If we see a duplicate commit/reveal message, we return an error.
	if seen[msgInfo] {
		return sdk.Context{}, errorsmod.Wrapf(sdkerrors.ErrInvalidRequest, "duplicate commit or reveal message detected: %s", msgInfo)
	}
	seen[msgInfo] = true
}
```

Thomasvdam (approved):

app/ante.go
hacheigriega: Yeah this is certainly cleaner. Updated to this and also added a test involving "nil commit messages" to address an audit finding.

PR #529, opened 2025-03-19
Prevent frontrunning data proxy registration

Thomasvdam:
## Motivation

Noone like front running unless it means they're only paying for your gas.

## Explanation of Changes

Require the signer address to be part of the signed payload.

## Testing

Updated test vectors with data from https://github.com/sedaprotocol/seda-data-proxy/pull/41

## Related PRs and Issues

Data proxy PR: https://github.com/sedaprotocol/seda-data-proxy/pull/41

### DIFF
--- a/x/data-proxy/keeper/msg_server.go
+++ b/x/data-proxy/keeper/msg_server.go
@@ -60,14 +60,20 @@ func (m msgServer) RegisterDataProxy(goCtx context.Context, msg *types.MsgRegist
 	}
 
 	feeBytes := []byte(msg.Fee.String())
+	adminAddressBytes := []byte(msg.AdminAddress)
 	payoutAddressBytes := []byte(msg.PayoutAddress)
 	memoBytes := []byte(msg.Memo)
+	chainIDBytes := []byte(ctx.ChainID())
 
-	payload := make([]byte, 0, len(feeBytes)+len(payoutAddressBytes)+len(memoBytes))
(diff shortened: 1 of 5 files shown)


x/data-proxy/keeper/msg_server.go
```diff
@@ -60,11 +60,13 @@ func (m msgServer) RegisterDataProxy(goCtx context.Context, msg *types.MsgRegist
 	}
 
 	feeBytes := []byte(msg.Fee.String())
+	adminAddressBytes := []byte(msg.AdminAddress)
 	payoutAddressBytes := []byte(msg.PayoutAddress)
 	memoBytes := []byte(msg.Memo)
 
 	payload := make([]byte, 0, len(feeBytes)+len(payoutAddressBytes)+len(memoBytes))
```
hacheigriega, line 67: Perhaps we can change the capacity?

hacheigriega: Not sure how to accomplish that without making the code dirty..

Thomasvdam: Good catch! I might just move the sum out but it'll still be kinda dirty :p

Thomasvdam: I also added memo length validation to the `ValidateBasic` methods.

hacheigriega: LGTM!

PR #531, opened 2025-03-21
fix(abci): handle invalid vote extensions in late votes

Thomasvdam:
## Motivation

Since votes that arrive before the commit timeout but after the block has been delivered to the application are not verified with the verifyVoteExtension handler we need to account for potentially invalid votes in the prepareProposalHandler. Since we at least are guaranteed to receive verified votes for over 2/3rds of voting power we can simply ignore any vote that was invalid.

## Explanation of Changes

Most of the VoteExtension code is taken directly from skip-mev/connect https://github.com/skip-mev/connect/blob/main/abci/ve/vote_extension.go, which in turn is taken almost directly from the SDK https://github.com/cosmos/cosmos-sdk/blob/v0.50.13/baseapp/abci_utils.go.

## Testing

Refactored the tests significantly to make every individual test a little more clear, since the malicious proposer scenario adds so many variations that it became unwieldy with the testcases approach.

I think it can be cleaner though, so comments very much welcome. :) 

## Related PRs and Issues

N.A.

### DIFF
--- a/app/abci/handlers.go
+++ b/app/abci/handlers.go
@@ -14,7 +14,6 @@ import (
 	addresscodec "cosmossdk.io/core/address"
 	"cosmossdk.io/log"
 
-	"github.com/cosmos/cosmos-sdk/baseapp"
 	sdk "github.com/cosmos/cosmos-sdk/types"
 
 	"github.com/sedaprotocol/seda-chain/app/utils"
@@ -162,7 +161,7 @@ func (h *Handlers) VerifyVoteExtensionHandler() sdk.VerifyVoteExtensionHandler {
 		err = h.verifyBatchSignatures(ctx, batch.BatchNumber, batch.BatchId, req.VoteExtension, req.ValidatorAddress)
 		if err != nil {
 			h.logger.Error("failed to verify batch signature", "req", req, "err", err)
-			return nil, err
+			return &abcitypes.ResponseVerifyVoteExtension{Status: abcitypes.ResponseVerifyVoteExtension_REJECT}, err
 		}
 
 		h.logger.Debug(
@@ -182,7 +181,7 @@ func (h *Handlers) PrepareProposalHandler() sdk.PrepareProposalHandler {
 		// Check if there is a batch whose signatures must be collected
 		// at this block height.
 		var collectSigs bool
-		_, err := h.batchingKeeper.GetBatchForHeight(ctx, ctx.BlockHeight()+BlockOffsetCollectPhase)
+		batch, err := h.batchingKeeper.GetBatchForHeight(ctx, ctx.BlockHeight()+BlockOffsetCollectPhase)
 		if err != nil {
 			if !errors.Is(err, collections.ErrNotFound) {
 				return nil, err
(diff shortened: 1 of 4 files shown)


app/abci/utils.go
```diff
@@ -0,0 +1,232 @@
...
+	"cosmossdk.io/core/comet"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+)
+
+// VoteExtensionsEnabled determines if vote extensions are enabled for the current block. If
+// vote extensions are enabled at height h, then a proposer will receive vote extensions
+// in height h+1. This is primarily utilized by any module that needs to make state changes
+// based on whether they were included in a proposal.
+func VoteExtensionsEnabled(ctx sdk.Context) bool {
```
hacheigriega, line 26: `IsVoteExtensionsEnabled()`?

app/abci/utils.go
```diff
@@ -0,0 +1,232 @@
...
+				return fmt.Errorf("non-commit vote extension signature present; validator addr %s",
+					vote.Validator.String(),
+				)
+			}
+		} else { // vote extensions disabled
+			if len(vote.VoteExtension) != 0 {
+				return fmt.Errorf("vote extension present but extensions disabled; validator addr %s",
+					vote.Validator.String(),
+				)
+			}
```
hacheigriega, line 113: Can this cause an issue to a new validator since he or she would not be expected to sign the first batch?

app/abci/utils.go
```diff
@@ -0,0 +1,232 @@
...
+
+	if extensionsEnabled {
+		// If the sum of the voting power has not reached (2/3 + 1) we need to error.
+		if requiredVP := ((totalVP * 2) / 3) + 1; sumVP < requiredVP {
+			return fmt.Errorf(
+				"insufficient cumulative voting power received to verify vote extensions; got: %d, expected: >=%d",
+				sumVP, requiredVP,
+			)
+		}
+	}
```
hacheigriega, line 175: Why do we check this only when the extension is enabled?

app/abci/utils.go
```diff
@@ -0,0 +1,232 @@
...
+		// If the validator does not have a valid public key, we skip the signature verification logic but still include
+		// the validator's voting power in the total voting power. The app may have pruned the validator's public key
+		// from the store, but comet considered the validator as active and included them in the commit since there
+		// is a 1 block delay between the validator set update on the app and comet.
+		sumVP += vote.Validator.Power
+		valConsAddr := sdk.ConsAddress(vote.Validator.Address)
+		pubKeyProto, err := valStore.GetPubKeyByConsAddr(ctx, valConsAddr)
+		if err != nil {
+			continue
+		}
```
hacheigriega, line 138: I'm unsure why 
(1) we add `sumVP` here when the signature verification hasn't happened
(2) the default SDK doesn't account for the scenario described in the comment.

app/abci/utils.go
```diff
@@ -0,0 +1,232 @@
...
+
+	for _, vote := range extCommit.Votes {
+		totalVP += vote.Validator.Power
+
+		if extensionsEnabled {
+			if vote.BlockIdFlag == cmtproto.BlockIDFlagCommit && len(vote.ExtensionSignature) == 0 {
+				return fmt.Errorf("vote extension signature is missing; validator addr %s",
+					vote.Validator.String(),
+				)
+			}
```
hacheigriega, line 97: Could this be an issue for a new active validator who doesn't sign the first batch?

app/abci/handlers.go
```diff
@@ -192,8 +191,26 @@ func (h *Handlers) PrepareProposalHandler() sdk.PrepareProposalHandler {
 		}
 
 		var injection []byte
-		if req.Height > ctx.ConsensusParams().Abci.VoteExtensionsEnableHeight && collectSigs {
-			err := baseapp.ValidateVoteExtensions(ctx, h.stakingKeeper, req.Height, ctx.ChainID(), req.LocalLastCommit)
+		if VoteExtensionsEnabled(ctx) && collectSigs {
+			for i, vote := range req.LocalLastCommit.Votes {
+				// Verify the signatures since they're not guaranteed to have passed VerifyVoteExtension.
+				if err := h.verifyBatchSignatures(ctx, batch.BatchNumber, batch.BatchId, vote.VoteExtension, vote.Validator.Address); err != nil {
```
hacheigriega, line 197: Shouldn't we also check if the voter was in the previous batch's validator tree?

hacheigriega: Ah, I think I was mistaken. If a validator joins at height `H`, it would not be in the local last commit at height `H+1`.

app/abci/utils.go
Thomasvdam: No idea, I took these functions from skip. I don't think it matters much since this function will only be called when they're active so might as well remove it :)

app/abci/utils.go
Thomasvdam: I'm also not sure why, I think we can stick more closely to the default implementation.

PR #535, opened 2025-03-24
Limit the amount of data proxy fee updates per block

Thomasvdam:
Targeting other PR for easier merging.

## Motivation

As discussed it should be super unlikely that any number of data proxies schedule their update for the same block, and even if a large number needs to be coordinated it shouldn't matter that they need to be spread out over a few blocks.

## Explanation of Changes

The value 25 was randomly decided so I'm open to other values. I went with this since it's low enough that iterating doesn't matter but high enough to allow for some random collisions.

## Testing

Added a unit test to verify that scheduling an update isn't possible once the max is reached, but becomes possible again once an update is rescheduled.

## Related PRs and Issues

N.A.

### DIFF
--- a/x/data-proxy/keeper/keeper.go
+++ b/x/data-proxy/keeper/keeper.go
@@ -71,12 +71,22 @@ func (k Keeper) RemoveFeeUpdate(ctx sdk.Context, height int64, pubKey []byte) er
 	return k.feeUpdateQueue.Remove(ctx, collections.Join(height, pubKey))
 }
 
-func (k Keeper) processProxyFeeUpdate(ctx sdk.Context, pubKeyBytes []byte, proxyConfig types.ProxyConfig, newFee *sdk.Coin, updateDelay uint32) (int64, error) {
+func (k Keeper) scheduleFeeUpdate(ctx sdk.Context, pubKeyBytes []byte, proxyConfig types.ProxyConfig, newFee *sdk.Coin, updateDelay uint32) (int64, error) {
 	// Determine update height
 	updateHeight := ctx.BlockHeight() + int64(updateDelay)
 	feeUpdate := &types.FeeUpdate{
 		NewFee:       newFee,
 		UpdateHeight: updateHeight,
+	}
+
+	// Check if the max updates per block is reached
+	keys, err := k.getFeeUpdateKeys(ctx, updateHeight)
+	if err != nil {
+		return 0, err
+	}
+
+	if len(keys) >= types.MaxUpdatesPerBlock {
+		return 0, types.ErrMaxUpdatesReached
 	}
 
 	// Delete previous pending update, if applicable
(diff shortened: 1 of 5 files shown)


x/data-proxy/keeper/keeper.go
```diff
@@ -79,6 +79,16 @@ func (k Keeper) processProxyFeeUpdate(ctx sdk.Context, pubKeyBytes []byte, proxy
...
+	if err != nil {
+		return 0, err
+	}
+
+	if len(keys) >= types.MaxUpdatesPerBlock {
+		return 0, types.ErrMaxUpdatesReached
+	}
+
 	// Delete previous pending update, if applicable
 	if proxyConfig.FeeUpdate != nil {
```
hacheigriega, line 93: Should this case be considered separately at the earlier stage if our intention is to always allow updating an existing  fee update?

x/data-proxy/keeper/keeper.go
```diff
@@ -79,6 +79,16 @@ func (k Keeper) processProxyFeeUpdate(ctx sdk.Context, pubKeyBytes []byte, proxy
 		UpdateHeight: updateHeight,
 	}
```
hacheigriega, line 81: By the way, I think this function should be renamed to `scheduleFeeUpdate()`, right?

x/data-proxy/keeper/keeper.go
Thomasvdam: I think we want to only allow rescheduling if the new update height is not already full, otherwise an attacker could still fill one height with a lot of updates. Or am I misunderstanding your comment?

hacheigriega: I just thought rescheduling could be allowed because it wouldn't create a new entry.

Thomasvdam: I think I'm missing something :D 

How I understand the scenario:
1. For block 100 we have 25 (max) updates scheduled.
2. Proxy A schedules an update for block 99, which now has 1 update scheduled.
3. Proxy A reschedules from 99 to 100, which would delete the entry at 99 and create one at 100 leaving us with 26 scheduled for block 100.

So in reality the order won't matter since the deletion of the update on 99 should revert once we error upon 'discovering' that there are already 25 updates scheduled for 100 leaving us with the same outcome. I just find the flow a little more natural to first check if we can 'reschedule' before deleting the previous entry.

hacheigriega: Ahh thanks for explaining. I was assuming rescheduling at the same block height as before.
In an unlikely scenario when the reschedule happens at the same height as before, I guess this will prevent an update even though the max won't be exceeded? I guess we can ignore this scenario, though.

Thomasvdam: Aaah yes, in that scenario it'll fail but if I'm putting on my contractor cap "it's not a bug, it's a feature! we're saving the user Txs fees" 😂

hacheigriega (approved):

PR #536, opened 2025-03-25
Data proxy registration fee

hacheigriega:
### Explanation

We charge a fixed, gov-controlled fee for registration of a data proxy to prevent spamming of data proxy registration transactions.


x/data-proxy/types/params.go
```diff
@@ -1,18 +1,28 @@
...
 )
 
 const (
+	MaxUpdatesPerBlock       int    = 25
 	DefaultMinFeeUpdateDelay uint32 = 86400 // Roughly 1 week with a ~7 sec block time
 	LowestFeeUpdateDelay     uint32 = 1
 )
 
+var (
+	DefaultRegistrationFee = math.NewIntWithDecimal(10, 18) // (10)*10^(18) aseda
```
Thomasvdam, line 18: I thought we discussed 50 as the initial fee?

Thomasvdam (commented):
Maybe we should update the registration unit tests to check that the fee was burnt and a testcase for an account that has insufficient funds?

x/data-proxy/types/params.go
```diff
@@ -10,10 +15,13 @@ const (
 	LowestFeeUpdateDelay     uint32 = 1
 )
 
+var DefaultRegistrationFee = math.NewIntWithDecimal(50, 18) // (10)*10^(18) aseda
```
Thomasvdam, line 18: The comment is outdated 😬

Thomasvdam (approved):

x/data-proxy/types/params.go
hacheigriega: ah thanks

PR #538, opened 2025-03-26
Data proxy and executor pubkey sorting

Thomasvdam:
## Motivation

We want to prevent executors/proxies with keys that sort favourably from obtaining rewards more often in the unlikely case where there is not enough aseda in escrow to reward all the participants.

## Explanation of Changes

I opted to use the bytes of the string representation so we don't have to deal with decoding the strings as hex.

We use the data request id and the block height at which the request is tallied as entropy to add together with the public key to produce a hash which we use for sorting.

## Testing

Added unit tests for the 

## Related PRs and Issues

N.A.

### DIFF
--- a/x/tally/types/hash_sort.go
+++ b/x/tally/types/hash_sort.go
@@ -1,0 +1,45 @@
+package types
+
+import (
+	"bytes"
+	"crypto/sha256"
+	"slices"
+)
+
+type HashSortable interface {
+	GetSortKey() []byte
+}
+
+func HashSort[T HashSortable](originalItems []T, entropy []byte) []T {
+	type sortItem struct {
+		sortKey  []byte
+		original T
+	}
+	totalItems := len(originalItems)
+
+	sortItems := make([]sortItem, totalItems)
+
+	hasher := sha256.New()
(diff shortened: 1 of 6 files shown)


hacheigriega (approved):
🔥

x/tally/types/hash_sort.go
```diff
@@ -0,0 +1,47 @@
...
+	sortItems := make([]sortItem, totalItems)
+
+	hasher := sha256.New()
+	for i := range totalItems {
+		hasher.Reset()
+
+		hasher.Write(originalItems[i].GetSortKey())
+		for _, entropy := range entropy {
+			hasher.Write(entropy)
+		}
```
hacheigriega, line 29: It's confusing that the iterator and the iteration list have the same name here.

hacheigriega: Why are we iterating through entropy here?

hacheigriega (changes requested):

gluax (approved):

mariocao (approved):

hacheigriega (approved):

PR #541, opened 2025-03-27
Replace standard deviation filter with MAD filter

hacheigriega:
## Motivation

The standard deviation filter was vulnerable to an attack where a malicious executor reports an exorbitant outlier.
Suppose, for instance, the following scenario devised by @Thomasvdam 

Suppose executor reports follow a bi-modal distribution:
```
1000, 1000, 1000, 5000, 5300, 5400, 5500
```
Assuming the sigma multiplier 1, with the standard deviation at 2303 and the mean at 3457, the first three executors will be classified as outliers. The outcome of standard deviation filter is "No consensus."

Here, a malicious executor can change his or her report to an exorbitant value (such as 115792089237316195423570985008687907853269) to change the consensus outcome to true.

[The MAD filter](https://en.wikipedia.org/wiki/Median_absolute_deviation) is a little more robust against such attacks because an outlier report from a minority group in the bi-modal distribution cannot change the consensus outcome.
For example,
```
1000, 1000, 115792089237316195423570985008687907853269, 5000, 5300, 5400, 5500
```
results in the same outcome as the original case.

## Explanation of Changes

The MAD filter replaces the standard deviation filter.
Also removed the fuzz test for the standard deviation filter since it wasn't testing anything important.

## Testing

Corrected the existing tests and added a few cases.


mariocao (approved):
💃

x/tally/types/filters.go
```diff
@@ -266,8 +256,44 @@ func detectOutliersBigInt(dataList []string, sigmaMultiplier SigmaMultiplier, er
 
 // isWithinMaxDev returns true if the given number is within the given
 // deviation amount from the mean.
```
gjaldon, line 258: Doc still states mean instead of median. :)

PR #544, opened 2025-04-01
Upgrade handler for v1.0.0

hacheigriega:
## Explanation of Changes

Upgrade handler named v1.0.0 stores and instantiates the Core Contract, registers its address, and enables vote extension at the upgrade height + 1. 

For the current plan for mainnet upgrade, see https://www.notion.so/sedaprotocol/Upgrading-1afa68d575ca80a081cec7898fcac04c?pvs=4#1c8a68d575ca806da281dc5432d23a1e

## Testing

Manually tested using the modified state of a mainnet export.


app/upgrades/mainnet/v1/constants.go
```diff
@@ -30,18 +33,57 @@ var Upgrade = upgrades.Upgrade{
...
+		ctx := sdk.UnwrapSDKContext(context)
+
+		// Run module migrations.
+		migrations, err := mm.RunMigrations(ctx, configurator, fromVM)
+		if err != nil {
+			return nil, err
+		}
+
+		// Additional upgrade logic for instantiating the Core Contract.
+		securityGroupAddr := "seda1afk9zr2hn2jsac63h4hm60vl9z3e5u69gndzf7c99cqge3vzwjzs026662"
```
mariocao, line 54: Just double checking, security group addresses are the same in testnet and mainnet, right?

mariocao (approved):

app/upgrades/mainnet/v1/constants.go
```diff
@@ -30,18 +33,57 @@ var Upgrade = upgrades.Upgrade{
...
+			return nil, err
+		}
+
+		codeID, _, err := keepers.WasmContractKeeper.Create(ctx, adminAddr, wasmCode, nil)
+		if err != nil {
+			return nil, err
+		}
+		contractAddr, _, err := keepers.WasmContractKeeper.Instantiate(
+			ctx, codeID, adminAddr, adminAddr,
+			[]byte(fmt.Sprintf(`{"token":"aseda", "owner": "%s", "chain_id":"%s"}`, securityGroupAddr, ctx.ChainID())),
```
Thomasvdam, line 66: Good reminder to think about the default values we want to set on the core contract for staking config and DR timeouts

app/upgrades/mainnet/v1/constants.go
hacheigriega: Yes. They have different admins but they have the same policy address.

app/upgrades/mainnet/v1/constants.go
hacheigriega: Will create a checklist in the Notion doc!

PR #545, opened 2025-04-03
Refactor reveal sorting in tally

hacheigriega:
## Explanation of Changes

PR #538 introduced sorting data proxy and executor payout lists with entropy in order to ensure some level of fairness when the contract goes through the lists in order until the fees are exhausted. This change, however, did not address the scenario where divergent gas reports involve multiple lowest reporters. To address both issues, this PR moves the executor sorting with entropy from the payout message construction phase to the beginning of tally process of a given request.

### DIFF
--- a/x/tally/keeper/filter.go
+++ b/x/tally/keeper/filter.go
@@ -13,9 +13,13 @@ const (
 	filterTypeMAD  byte = 0x02
 )
 
+// FilterResult is the result of filtering.
 type FilterResult struct {
-	Errors       []bool   // i-th item is true if i-th reveal is non-zero exit or corrupt
-	Outliers     []bool   // i-th item is non-zero if i-th reveal is an outlier
+	// Ordered results sorted by executor with entropy
+	Executors []string // list of executor identifiers
+	Errors    []bool   // i-th item is true if i-th reveal is non-zero exit or corrupt
+	Outliers  []bool   // i-th item is non-zero if i-th reveal is an outlier
+	// Consensus results
 	Consensus    bool     // whether consensus (either in data or in error) is reached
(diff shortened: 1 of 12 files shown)


x/tally/keeper/filter.go
```diff
@@ -13,7 +13,9 @@ const (
 	filterTypeMAD  byte = 0x02
 )
 
+// FilterResult is the result of filtering.
 type FilterResult struct {
+	Executors    []string // list of executors
```
mariocao, line 18: If I understood the code...this list of executors is already "shuffled" or sanitize (with sort)
```suggestion
	Executors    []string // list of sorted/sanitized executors
```

x/tally/keeper/filter_and_tally_test.go
```diff
@@ -340,7 +353,68 @@ func TestExecutorPayout(t *testing.T) {
 			},
 		},
 		{
-			name:            "Divergent gas reporting without shares (lowest_report*2 < median_report)",
+			name:            "Divergent gas reporting with multiple lows (low*2 > median)",
+			tallyInputAsHex: "00",
+			reveals: map[string]types.RevealBody{
+				"lizard":  {ExitCode: 0, Reveal: `{"result": {"text": "A"}}`, GasUsed: 28000},
+				"bonobo":  {ExitCode: 0, Reveal: `{"result": {"text": "A"}}`, GasUsed: 30000},
```
mariocao, line 360: 🐵

mariocao (approved):
lgtm!

x/tally/keeper/filter_and_tally_test.go
hacheigriega: 🦎

gluax (approved):

PR #547, opened 2025-04-14
Custom app config for SEDA key file path

hacheigriega:
## Explanation of Changes
We have been using the same directory as the private validator key file for storing and finding the SEDA key file since this path is provided by the CometBFT configuration. Since the SEDA key file is used by the application, this PR creates a new configuration to the `app.toml` file. The default location for the key file remains the same at `$NODE_HOME/config/seda_keys.json`.

This configuration value can be queried with the following command:

```
$ sedad config get app seda.seda_key_file             
"config/seda_keys.json"
```


hacheigriega:
Sorry, fixing e2e right now..

Thomasvdam (approved):

PR #548, opened 2025-04-16
Migration of app.toml and additions to SEDA config

hacheigriega:
## Explanation of Changes

- Pre-upgrade command for migrating `app.toml`.
- Add app configurations for whether to enable SEDA signer and whether to allow use of unencrypted SEDA keys. Also inject Viper configuration in PreRun of the start command so that the app can know when the node is starting. This config cannot be changed by user.


hacheigriega:
Confused why the gentx command is suddenly failing in interchain tests. Will fix tomorrow

cmd/sedad/cmd/config.go
```diff
@@ -81,3 +79,44 @@ func initTendermintConfig() *tmcfg.Config {
...
+
+	serverconfig.SetConfigTemplate(serverconfig.DefaultConfigTemplate)
+	oldConfig := serverconfig.DefaultConfig()
+	err := serverCtx.Viper.Unmarshal(oldConfig)
+	if err != nil {
+		return fmt.Errorf("failed to parse %s: %w", appConfigPath, err)
+	}
+
+	defaultSedaConfig := utils.DefaultSEDAConfig()
+	defaultSedaConfig.AllowUnencryptedSEDAKeys = true
```
Thomasvdam, line 114: I suppose we're setting this to true here to a node doesn't immediately panic upon updating as it's unlikely operators will have set up the encryption key beforehand? Maybe it's good to add brief comment to explain we set it here rather than modifying the default?

cmd/sedad/cmd/config.go
```diff
@@ -81,3 +79,44 @@ func initTendermintConfig() *tmcfg.Config {
 
 	return cfg
 }
+
+func preUpgradeCmd() *cobra.Command {
```
Thomasvdam, line 83: Do we have to worry about this command with future upgrades? Given that I don't see versions I guess it always runs.

Given that we set the `AllowUnencryptedSEDAKeys` to true in the migration this could potentially alter an operators config if they've set it to false.

Thomasvdam (approved):

cmd/sedad/cmd/config.go
hacheigriega: Added the comment. We now also require the values of `--key-file-no-encryption` flag and `seda.allow-unencrypted-seda-keys` app config to match when creating a key file.

cmd/sedad/cmd/config.go
hacheigriega: Yes good point.. Cosmovisor always runs it before applying an upgrade, so this will have to be removed in future upgrades.

Thomasvdam: Or we add a version to the config file and check that to decide which migration to execute?

hacheigriega: I addressed the issue by just checking if seda config already exists here https://github.com/sedaprotocol/seda-chain/pull/549 I'd appreciate your review there 🙏

PR #549, opened 2025-04-22
Minor refactor of SEDA Keys and config

hacheigriega:
-  Do not read environment variable for encryption key when generating SEDA Keys. 
   - From my experience, using the environment variable value as an encryption key when generating SEDA Keys was prone to cause confusion. Now to use a custom encryption key, you have to pass the custom encryption flag and provide the key when prompted. 
- In pre-upgrade, return success without migrating if "seda" app config already exists.
- Fix local multi validator setup script.


Thomasvdam:
I'm not sure how to fix the interchain tests... This was part of the reason I had the undocumented env variable as it's seemingly really hard to customise the `gentx` command :(

Thomasvdam (approved):

PR #552, opened 2025-04-24
Improve SEDA signer configuration logic

hacheigriega:
## Explanation of Changes

- Panic at startup if signer loading fails when it is enabled in config.
- Fix pre-upgrade logic.
- ExtendVote should not fail even if signer reload due to mismatch fails.

### DIFF
--- a/cmd/sedad/cmd/config.go
+++ b/cmd/sedad/cmd/config.go
@@ -101,7 +101,7 @@ func preUpgradeCmd() *cobra.Command {
 
 func migrateAppConfig(serverCtx *server.Context, rootDir string) error {
 	value := serverCtx.Viper.Get("seda")
-	if value == nil {
+	if value != nil {
 		return nil // seda config already exists
 	}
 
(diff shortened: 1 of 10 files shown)


hacheigriega:
Interchain test again :/ Fixing it now

cmd/sedad/cmd/config.go
```diff
@@ -101,7 +101,7 @@ func preUpgradeCmd() *cobra.Command {
 
 func migrateAppConfig(serverCtx *server.Context, rootDir string) error {
 	value := serverCtx.Viper.Get("seda")
-	if value == nil {
+	if value != nil {
```
Thomasvdam, line 104: Whoops :D

x/pubkey/types/params.go
```diff
@@ -5,7 +5,7 @@ import (
 )
 
 const (
-	DefaultActivationBlockDelay       = 11520 // roughly 1 day
```
Thomasvdam, line 8: Do we want to change this?

Thomasvdam (approved):

x/pubkey/types/params.go
hacheigriega: For testing yes. For mainnet release we would have to go over the default parameters again. Added that to the v1.0.0 release checklist on Notion.

cmd/sedad/cmd/config.go
hacheigriega: Yeah 😓

PR #562, opened 2025-05-15
Support setting SEDA key file encryption key through a file

hacheigriega:
## Motivation

There were requests from validators to support configuring the encryption key for a SEDA key file through a file in addition to an environment variable. This PR adds a new environment variable `FILE_SEDA_KEYS_ENCRYPTION_KEY` whose value, if present, would indicate a path to the file that contains an encryption key to the SEDA key file.

Also adds a dependency list to the pull request template.

### DIFF
--- a/app/utils/seda_keys.go
+++ b/app/utils/seda_keys.go
@@ -24,15 +24,31 @@ import (
 )
 
 const (
-	// SEDAKeyEncryptionKeyEnvVar is the environment variable that should contain the SEDA key encryption key.
+	// SEDAKeyEncryptionKeyEnvVar is the environment variable that contains
+	// the encryption key to the SEDA key file.
 	SEDAKeyEncryptionKeyEnvVar = "SEDA_KEYS_ENCRYPTION_KEY"
+	// SEDAKeyEncryptionKeyFile is the environment variable that contains the
+	// path to the file that contains the encryption key to the SEDA key file.
+	SEDAKeyEncryptionKeyFile = "FILE__SEDA_KEYS_ENCRYPTION_KEY"
 )
 
-// ReadSEDAKeyEncryptionKeyFromEnv reads the SEDA key encryption key from
-// the environment variable. Returns an empty string if the environment
-// variable is not set.
+// ReadSEDAKeyEncryptionKeyFromEnv attempts to read an encryption key to the
+// SEDA key file from the environment variable. It returns an empty string if
+// no encryption key is found. We also allow users to specify a file that
+// contains an encryption key through an environment variable.
 func ReadSEDAKeyEncryptionKeyFromEnv() string {
-	return os.Getenv(SEDAKeyEncryptionKeyEnvVar)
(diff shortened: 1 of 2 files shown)


app/utils/seda_keys.go
```diff
@@ -24,15 +24,29 @@ import (
 )
 
 const (
-	// SEDAKeyEncryptionKeyEnvVar is the environment variable that should contain the SEDA key encryption key.
+	// SEDAKeyEncryptionKeyEnvVar is the environment variable that contains
+	// the encryption key to the SEDA key file.
 	SEDAKeyEncryptionKeyEnvVar = "SEDA_KEYS_ENCRYPTION_KEY"
+	// SEDAKeyEncryptionKeyFile is the file that contains the encryption key
+	// to the SEDA key file.
+	SEDAKeyEncryptionKeyFile = "FILE_SEDA_KEYS_ENCRYPTION_KEY"
```
Thomasvdam, line 32: The Discord message mentions `FILE__{ENV_NAME}` (2 underscores).

app/utils/seda_keys.go
```diff
@@ -24,15 +24,29 @@ import (
...
 
-// ReadSEDAKeyEncryptionKeyFromEnv reads the SEDA key encryption key from
-// the environment variable. Returns an empty string if the environment
-// variable is not set.
+// ReadSEDAKeyEncryptionKeyFromEnv attempts to read SEDA keys encryption key
+// from the environment variable and then from the file. It returns an empty
+// string if neither of the two attempts succeeds.
 func ReadSEDAKeyEncryptionKeyFromEnv() string {
-	return os.Getenv(SEDAKeyEncryptionKeyEnvVar)
+	key := os.Getenv(SEDAKeyEncryptionKeyEnvVar)
```
Thomasvdam, line 39: The way I'm reading how this flow should work (although I couldn't really find any documentation on it) is as follows:

1. Check for the `FILE__{ENV_NAME}` variable.
2. If present, assume the value is a filepath.
3. Load the contents of the filepath as a string and assign it to whatever `{ENV_NAME}` would normally have been.

It doesn't really say much about what should happen when both `FILE__{ENV_NAME}` and `{ENV_NAME}` are specified. I think we should probably use `FILE__{ENV_NAME}` as leading based on what I saw where people would use this flow in Docker images where on startup the container basically does:

1. Loop over all env variables with the `FILE__` prefix
2. Read the files found at the paths in those env vars
3. Assign the file contents to an env var with the original name minus the `FILE__` prefix.

Thomasvdam (changes requested):

app/utils/seda_keys.go
Thomasvdam: For future reference this is most likely the script that this suggestion was based on: https://github.com/linuxserver/docker-baseimage-alpine/blob/master/root/etc/s6-overlay/s6-rc.d/init-envfile/run

Thomasvdam (approved):

PR #566, opened 2025-05-27
Improve vote extension log readability

hacheigriega:


app/abci/handlers.go
```diff
@@ -152,9 +153,9 @@ func (h *Handlers) VerifyVoteExtensionHandler() sdk.VerifyVoteExtensionHandler {
 				}
 				h.logger.Error(
 					"received vote extension even though we're skipping batching",
-					"request", req.ValidatorAddress,
+					"validator_address", hex.EncodeToString(req.ValidatorAddress),
```
Thomasvdam, line 156: Is the `valoper....` format an option for the log statements?

hacheigriega: That'd be more helpful huh. But I didn't want to bring more code changes just for logging and as long as we have a hex address, we can use `sedad debug addr` command to get the bech addresses

Thomasvdam (approved):

PR #574, opened 2025-06-30
Update core contract instantiate msg

hacheigriega:
- Removes core contract parameter values from instantiate msg so that the contract parameters are initialized to the default values defined in the contract.
- Telemetry sets a gauge under key `seda_tally_end_block_dr_flow_halt` to 1 if DR flow halts. It's set back to 0 when `ProcessTallies()` is executed successfully.


gluax (commented):
Telemetry changes look good to me.
I don't remember if there was a reason we also wanted to hard code the configs for the contract, so I'll let someones else chime in there.

Thomasvdam (approved):
I think having 1 less place to keep those values in sync with is a good thing :)

PR #579, opened 2025-07-08
chore: adjust block times in config init

gluax:
## Motivation

Our new avg block time is `8s` so we need to update things

## Explanation of Changes

- make a constant and use that for the `TimoutPropose` and `TimeoutCommit`

I believe that only works for new configs- not existing ones right?
We also need to adjust `blocks_per_year` in the genesis config yes? But I'm really not sure the correct way to do that either as it would need to be in the upgrade proposal somehow right?

Not sure that I know enough about how the chain works to really do this :x

## Testing

Unit tests all still pass- but I didn't change much :sweat_smile:

## Dependency Chain
[data-proxy]:https://github.com/sedaprotocol/seda-data-proxy
[evm-contracts]:https://github.com/sedaprotocol/seda-evm-contracts
[sdk]: https://github.com/sedaprotocol/seda-sdk
[overlay-rs]: https://github.com/sedaprotocol/overlay-rs
[overlay-ts]: https://github.com/sedaprotocol/seda-overlay-ts
[solver]:https://github.com/sedaprotocol/seda-solver
[chain-indexer-plugin]: https://github.com/sedaprotocol/seda-chain
[explorer/indexer]: https://github.com/sedaprotocol/seda-explorer

Think about the changes (msgs, events, limits, etc.) made in this PR and see if you need to make an issue or a PR on the following repos.

- [ ] [data-proxy][data-proxy]
- [ ] [evm-contracts][evm-contracts]
- [ ] [sdk][sdk]
- [ ] [overlay-rs][overlay-rs]
- [ ] [overlay-ts][overlay-ts]
- [ ] [solver][solver]
- [ ] [chain-indexer-plugin][chain-indexer-plugin]
- [ ] [explorer/indexer][explorer/indexer]

## Related PRs and Issues

N/A


cmd/sedad/cmd/config.go
```diff
@@ -73,9 +75,9 @@ func initCometConfig() *cmtcfg.Config {
 	cfg.RPC.ListenAddress = "tcp://0.0.0.0:26657"
 
 	// Consensus Settings
-	cfg.Consensus.TimeoutPropose = time.Duration(7.5 * float64(time.Second))
+	cfg.Consensus.TimeoutPropose = time.Duration(blockTime)
 	cfg.Consensus.TimeoutProposeDelta = time.Duration(0)
-	cfg.Consensus.TimeoutCommit = time.Duration(7.5 * float64(time.Second))
+	cfg.Consensus.TimeoutCommit = time.Duration(blockTime)
```
hacheigriega, line 80: I think we should leave these values because this might change the block time?

hacheigriega:
Right. The `blocks_per_year` parameter of the mint module should probably be updated through another inflation parameter change proposal. These changes are totally independent from that.

Thomasvdam (approved):

PR #580, opened 2025-07-21
Cleanup temp directories from pre-instantiation

hacheigriega:
A temporary directory created from pre-instantiation of the application was not being properly clean up. As a result, every run of the application created a directory that never got cleaned up. 

The issue originates from the code borrowed from x/wasmd, which has since been fixed https://github.com/CosmWasm/wasmd/pull/2018. This PR applies this fix.

A minor note: This PR also follows the wasmd standard to use a temp directory as the node home for the pre-instantiation process. Before, we were simply creating a temporary wasm directory under the default node home for pre-instantiation.

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -71,6 +71,7 @@ import (
 
 	"github.com/cosmos/cosmos-sdk/baseapp"
 	"github.com/cosmos/cosmos-sdk/client"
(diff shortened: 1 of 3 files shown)


Thomasvdam (commented):
The code reads pretty similar to me so I guess it's some Golang nuance that the first method doesn't clean up properly?

Issue #595, opened 2025-08-21
✨ x/fast


Thomasvdam:
## ✨ Feature

A module enabling SEDA fast; programmable API access and offchain execution.

## Motivation

SEDA's current model allows it to process data requests in a decentralised, permissionless, and verifiable manner. This comes at the cost of speed as we need settlement time for commit/reveal phases and batch signature aggregation. SEDA Fast allows access to SEDA's programmability in a much faster, but more permissioned and trusted way.

Since Fast works with the exact same Oracle Programs as SEDA an integrating party can easily transition from Fast to vanilla SEDA should there be a stronger requirement for decentralisation. They could even combine SEDA core and SEDA Fast to do soft updates on a faster interval and only settle once the verifiable result from SEDA core comes in.

## Implementation

A new module which:
- Has an 'owner' address which can be transferred
- Maintains a 'credits' balance for users (IDs)
- Maintains a balance of spent credits
- Allows the owner to top up credits for a user
- Can receive Fast client reports and then emits events for the individual reports, transfers data proxy fees, and deducts credits from the users
- Maintains an allowlist of Fast clients
- Allows the owner to create new Fast clients
- Checks eligibility of a Fast client request to a data proxy
- Allows the owner to settle the spent credits (to be determined if this is a withdraw and/or burn)

PR #599, opened 2025-08-26
feat(x/sophon): add store, keeper methods, and genesis import/export

Thomasvdam:
## Motivation

Allow Sophon data to be persisted.

## Explanation of Changes

Knowledge of the collections and how they are set up should be limited to the `keeper.go` file.

We use the `sophon_id` for the SophonUser key to make public key rotation for the Sophon easier. Now only the SophonInfo has to be moved to a new store entry and we don't need to iterate over all the users to update them. Since most queries and transactions that use the SophonUser also require the SophonInfo this does not introduce unnecessary store access.

## Testing

Added a unit test that creates some state, exports it, imports the export, and verifies the imported state matches the state from before the export.

Tested starting the chain locally and that still works, but obviously nothing else in the Sophon module works as nothing is wired up/implemented for the messages and queries.

## Related PRs and Issues

Part-of: #595


gluax (approved):

x/sophon/types/keys.go
```diff
@@ -7,4 +7,10 @@ const (
 	StoreKey
 )
 
-var ParamsPrefix = collections.NewPrefix(0)
+var (
+	ParamsPrefix         = collections.NewPrefix(0)
+	SophonIDPrefix       = collections.NewPrefix(1)
```
hacheigriega, line 12: nitpick but Cosmos convention seems to be that standalone keys are named with the suffix "Key" instead of "Prefix." So `ParamsKey` and `SophonIDKey`.

x/sophon/types/sophon_info.go
```diff
@@ -2,8 +2,52 @@ package types
...
+		return fmt.Errorf("public key is empty")
+	}
+
+	if s.Balance.IsNil() {
+		return fmt.Errorf("balance is nil")
+	}
+
+	if s.UsedCredits.IsNil() {
+		return fmt.Errorf("used credits is nil")
+	}
```
hacheigriega, line 50: Maybe we want to check that they are `IsPositive()`? Not sure

x/sophon/types/sophon_user.go
```diff
@@ -5,5 +5,13 @@ import (
...
 
 func (s *SophonUser) ValidateBasic() error {
-	return fmt.Errorf("not implemented")
+	if len(s.UserId) == 0 {
+		return fmt.Errorf("user id is empty")
+	}
+
+	if s.Credits.IsNil() {
+		return fmt.Errorf("credits is nil")
+	}
```
hacheigriega, line 14: ditto

hacheigriega (approved):

PR #605, opened 2025-08-28
feat(x/sophon): add AddUser message

Thomasvdam:
## Motivation

Allow the admin to register new users. Optionally with initial credits.

## Explanation of Changes

Also be more consistent with the naming of the user_id, and make it a string as this is how we'll most likely deal with it in external systems too.

## Testing

Add tests for message validation and handling.

## Related PRs and Issues

Part-of: #595


x/sophon/types/tx.go
```diff
@@ -151,7 +151,53 @@ func (m *MsgCancelOwnershipTransfer) ValidateBasic() error {
...
+	if len(m.SophonPublicKey) == 0 {
+		return sdkerrors.ErrInvalidRequest.Wrap("empty public key")
+	}
+
+	_, err := hex.DecodeString(m.SophonPublicKey)
+	if err != nil {
+		return sdkerrors.ErrInvalidRequest.Wrapf("invalid hex in pubkey: %s", m.SophonPublicKey)
+	}
+
+	if len(m.UserId) == 0 {
```
hacheigriega, line 167: Shouldn't we add some additional constraints on the user ID?

hacheigriega (approved):

x/sophon/types/tx.go
Thomasvdam: Good one. I went with 66 as most of our public key identifiers are hex strings of that length.

PR #610, opened 2025-09-02
feat(x/sophon): add SubmitReport message

Thomasvdam:
## Motivation

Allow the Sophon to publish the reports, triggering credits rebalancing and data proxy payouts.

## Explanation of Changes

I made one obviously opinionated decision:
- An unregistered data proxy does not cause an error. The credits are only consumed from the user and there is no update of the Sophon's balance.
- An invalid payout address on the data proxy is handled the same way.

In both these scenarios we do emit an event to help offchain resolution.

## Testing

Add new validation and handler unit tests.

## Related PRs and Issues

Part-of: #595


x/sophon/keeper/msg_server_credits.go
```diff
@@ -88,6 +89,145 @@ func (m msgServer) SettleCredits(goCtx context.Context, msg *types.MsgSettleCred
...
+	}
+
+	denom, err := m.stakingKeeper.BondDenom(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	totalDataProxyCredits := math.NewInt(0)
+	// Transfer the data proxy credits to the data proxy.
+	for dataProxyPubKey, amount := range dataProxyCredits {
```
hacheigriega, line 165: Since map iteration is not deterministic, I think we would have to just pay the data proxies in the earlier loop without constructing `dataProxyCredits`. Or `dataProxyCredits` could be a slice that we sort before making the payouts.

hacheigriega (changes requested):

PR #615, opened 2025-09-05
x/core staking

hacheigriega:
## Explanation of Changes

This PR adds a store to x/core for stakers and adds a new message `Stake`. 

Note that the stake hash formula now does not include the Core Contract address.


x/core/keeper/msg_server.go
```diff
@@ -52,6 +55,100 @@ func (m msgServer) AddToAllowlist(goCtx context.Context, msg *types.MsgAddToAllo
...
+	}
+	publicKey, err := hex.DecodeString(msg.PublicKey)
+	if err != nil {
+		return nil, err
+	}
+	proof, err := hex.DecodeString(msg.Proof)
+	if err != nil {
+		return nil, err
+	}
+	_, err = vrf.NewK256VRF().Verify(publicKey, proof, hash)
```
gluax, line 86: So in the contract/rust-common we have one statically created instance of the VRF struct and then use it when necessary instead of creating a new one every time.

Not for this PR, and that might have mattered more for WASM world than not, but might be worth considering at some point since that type of math can be expensive to re-instantiate every time.

gluax (approved):

x/core/types/msg.go
```diff
@@ -0,0 +1,60 @@
...
+	allBytes = append(allBytes, []byte(coreContractAddr)...)
+	allBytes = append(allBytes, seqBytes...)
+
+	hasher := sha3.NewLegacyKeccak256()
+	hasher.Write(allBytes)
+	return hasher.Sum(nil), nil
+}
+
+func (m MsgStake) ComputeStakeHash(chainID string, sequenceNum uint64) ([]byte, error) {
+	memoBytes, err := hex.DecodeString(m.Memo)
```
hacheigriega, line 39: Not sure why the memo type is now treated as an hex instead of base64. Double check this

PR #623, opened 2025-09-16
Core Contract query shim and a few core queriers

hacheigriega:
## Explanation of Changes

- Added CLI endpoint for `add-to-allowlist` tx in x/core. Need to add the rest.
- "Shim" in x/wasm to redirect Core Contract queries to x/core and decode their
responses.
- New x/core queriers `StakerAndSeq`, `StakingConfig`, and `DataRequestConfig`.

## Testing

Tested in local environment with seda-overlay-ts

### DIFF
--- a/x/wasm/query_msgs.go
+++ b/x/wasm/query_msgs.go
@@ -1,0 +1,114 @@
+package wasm
+
+import (
+	"encoding/json"
+	"strconv"
+
+	"github.com/cosmos/cosmos-sdk/codec"
+
+	coretypes "github.com/sedaprotocol/seda-chain/x/core/types"
+)
+
+const (
+	QueryPathStakerAndSeq      = "/sedachain.core.v1.Query/StakerAndSeq"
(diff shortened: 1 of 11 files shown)


x/wasm/query_msgs.go
```diff
@@ -0,0 +1,114 @@
...
+type CoreContractQuery struct {
+	GetStakerAndSeq  *GetStakerAndSeq  `json:"get_staker_and_seq"`
+	GetStakingConfig *GetStakingConfig `json:"get_staking_config"`
+}
+
+type GetStakerAndSeq struct {
+	PublicKey string `json:"public_key"`
+}
+
+type GetStakerAndSeqResponse struct {
```
gluax, line 27: Can't we use the protobuf types for this instead of double creating structs?

`QueryStakerAndSeqResponse` from the protobuf looks similar to this one no?

gluax: ah nvm

Issue #593, opened 2025-08-14
🔧 Deployment preparation


hacheigriega:
## 🔧 Chore

For the core module to be prepared for deployment, the following aspects must be handled.

- [x] Genesis state with import / export support
- [x] Migration upgrade handler
- [ ] Module README
- [ ] Additional documentation
- [ ] Decide whether to remove wasm vm

PR #637, opened 2025-09-23
x/core GenesisState 

hacheigriega:
## Explanation of Changes

Add genesis state so we can import and export the core module state.

## Testing

Integration testing to simulate a lot of data requests followed by comparison of states before and after export and import. The test also exports states from two different instantiations of the app and compares the two exported genesis states.

## Related PRs and Issues

Related: #593

### DIFF
--- a/proto/sedachain/core/v1/genesis.proto
+++ b/proto/sedachain/core/v1/genesis.proto
@@ -10,11 +10,29 @@ option go_package = "github.com/sedaprotocol/seda-chain/x/core/types";
 // GenesisState defines core module's genesis state.
 message GenesisState {
   Params params = 1 [ (gogoproto.nullable) = false ];
-  bool paused = 2;
-  string owner = 3 [ (cosmos_proto.scalar) = "cosmos.AddressString" ];
-  string pending_owner = 4 [ (cosmos_proto.scalar) = "cosmos.AddressString" ];
-  uint64 committing_count = 5;
-  uint64 revealing_count = 6;
-  uint64 tallying_count = 7;
-  uint32 staker_count = 8;
+  string owner = 2 [ (cosmos_proto.scalar) = "cosmos.AddressString" ];
+  string pending_owner = 3 [ (cosmos_proto.scalar) = "cosmos.AddressString" ];
+  bool paused = 4;
+
+  // Staking-related states:
+  repeated string allowlist = 5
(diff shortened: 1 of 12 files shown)


proto/sedachain/core/v1/core.proto
```diff
@@ -112,16 +112,18 @@ enum DataRequestStatus {
 message RevealBody {
   // Hex-encoded ID of the data request
   string dr_i_d = 1 [ (gogoproto.jsontag) = "dr_id" ];
+  // Hex-encoded public key of the executor who has revealed this.
+  string executor = 2;
```
gluax, line 116: This is an addition to the `RevealBody` compared to what we used to have. Why is this necessary? It also means we have to update the hash, and reveal body size limits since it's another field...

proto/sedachain/core/v1/genesis.proto
```diff
@@ -10,11 +10,19 @@ option go_package = "github.com/sedaprotocol/seda-chain/x/core/types";
...
+  bool paused = 4;
+
+  // Staking-related states:
+  repeated string allowlist = 5
+      [ (cosmos_proto.scalar) = "cosmos.AddressString" ];
+  repeated Staker stakers = 6 [ (gogoproto.nullable) = false ];
+
+  // Data request-related states:
+  repeated DataRequest data_requests = 7 [ (gogoproto.nullable) = false ];
+  repeated RevealBody reveal_bodies = 8 [ (gogoproto.nullable) = false ];
```
gluax, line 24: Oh so we have to store these in the genesis file? Isn't that going to be expensive since that's read/writing to a JSON file? How come a store isn't enough for them?

gluax: Or is it just initialization purposes? Sorry I don't know much about how the genesis section really works yet :sweat_smile:

proto/sedachain/core/v1/core.proto
gluax: Maybe it's ok because we overwrite it and ignore any value given https://github.com/sedaprotocol/seda-chain/blob/96d991c7cd030cfbeefc32ecf2b26c6a3a2ea0fb/x/core/keeper/msg_server_dr.go#L349. @mariocao can you double-check to make sure I'm not missing anything security wise?

It would still affect gas costs, and RPC possibly performance.

gluax (approved):

PR #647, opened 2025-10-06
Fix stake counter overwrite bug in upgrade handler

hacheigriega:
Execute RunMigrations() before the custom migration logic to prevent RunMigrations's InitGenesis from overwriting the staker counter.

This made PostDataRequest fail after upgrade due to replication factor always exceeding the zero staker count.

### Testing

Local upgrade test and checking PostDataRequest goes through after upgrade.


hacheigriega:
Oops, didn't realize the test is broken. Fixing it now.

gluax (approved):

PR #650, opened 2025-10-08
Data request store refactor

hacheigriega:
## Motivation

Since proto encoding of maps is not deterministic, we remove maps `commits` and `reveals` from type `DataRequest` and instead store them separately.

## Explanation of Changes

Core module states:
```
allowlist:             0x00 | PublicKey              -> ()
stakers:               0x01 | PublicKey              -> Staker
staker_index_to_key:   0x02 | uint32                 -> PublicKey
staker_key_to_index:   0x03 | PublicKey              -> uint32
staker_count:          0x04                          -> uint32
data_requests:         0x05 | DR_ID                  -> DataRequest
data_request_indexing: 0x06
committing_count:      0x07 | DataRequestIndex       -> ()
revealing_count:       0x08 | DataRequestIndex       -> ()
tallying_count:        0x09 | DataRequestIndex       -> ()
commits:               0x0A | DR_ID | PublicKey      -> Commit
revealers:             0x0B | DR_ID | PublicKey      -> ()
reveal_bodies:         0x0C | DR_ID | PublicKey      -> RevealBody
timeout_queue:         0x0D | DR_ID | Timeout_Height -> ()
params:                0x0E                          -> Params
owner:                 0x0F
paused:                0x10
pending_owner:         0x11
```


x/core/keeper/data_request_tests/reveal_test.go
gluax, line 1: This is a breaking API change from the contract- Querying a Data Request should still return a Data Request object with the Reveals and Commits.

I think we should have two Protobuf versions of a Data Request. One for responses that have the commits and reveals and one for the state managment without them. Similar to how the contract did it.

gluax: Especially since these tests here are written as if by callers of the `GetDataRequestsByStatus` query- they(the overlay) won't have access to the keeper to do these checks and would instead have to do a secondary query. No bueno.

Issue #663, opened 2025-11-12
🔧 Prune batching module state


hacheigriega:
## 🔧 Chore

We want to prune batching module state, which is currently growing without a bound. 

Pruning batching module will be parameterized by the minimum number of batches to keep and the maximum number of batches to prune per block. The former parameter will be chosen to match the unbonding period, and the latter will be determined based on consideration of deletion computation time and time it will take to reach the minimum number of batches.

Note the batch pruning involves pruning batch data such as validator tree entries, data tree entries, and batch signatures. It remains a question how to go about pruning data results, which are not indexed by batch number.

PR #664, opened 2025-11-12
Pruning batches

hacheigriega:
## Explanation of Changes

Prune batches and their associated data at every block based on two new module parameters NumBatchesToKeep and MaxBatchPrunePerBlock. 

See #667 for pruning data results.

See the branch `hy/batch-prune-opt` for further optimization attempts.

## Testing

Added an integration test

## Related PRs and Issues

Related #663


Thomasvdam (commented):
If we clear the data tree entries how will we be able to figure out which data results can be pruned?

Oh wait we store results under the DR ID and height, it's difficult regardless :P

PR #667, opened 2025-11-20
Pruning data results

### LINKED ISSUE #663: 🔧 Prune batching module state

hacheigriega:
## Explanation of Changes

The challenge here is that there is no mapping to data result object from batch number or data result ID, which is only data available to us when we prune batches. Therefore, we simply go through `MaxDataResultsToCheckForPrune` (a module parameter) items in the store and delete those whose associated batches have been pruned. 

To prevent checking similar set of data results over multiple blocks, we use the hash of last commit hash as the starting point of this iteration. The iteration direction is also switched at every block.

Note this PR also includes some logic change of batching EndBlock to ensure that pruning is executed at every block and to catch errors from pruning functions without halting the chain.

## Testing

An integration test.

Planning to test with actual data exported from testnet or mainnet.

## Related PRs and Issues

Closes #663


x/batching/keeper/endblock_pruning.go
```diff
@@ -0,0 +1,150 @@
...
+	sdk "github.com/cosmos/cosmos-sdk/types"
+)
+
+func (k Keeper) PruneDataResults(ctx sdk.Context, maxDataResultsToCheckForPrune, lastRemovedBatchNum uint64) error {
+	if maxDataResultsToCheckForPrune == 0 || lastRemovedBatchNum == 0 {
+		k.Logger(ctx).Info("skip data result pruning", "max_data_results_to_check_for_prune", maxDataResultsToCheckForPrune, "last_removed_batch_num", lastRemovedBatchNum)
+		return nil
+	}
+
+	// Use hash of last commit hash as starting point of the range.
```
hacheigriega: note to self: remove this as this doesn't help

PR #673, opened 2025-11-28
Legacy data result pruning and basic pruning

### LINKED ISSUE #671: ♻️ Improve x/batching pruning
### Batch Pruning

Batch pruning was implemented in #664 and is complete.

TODO: Explain further

### Data Result Pruning and Simple Pruning

Data result pruning is being implemented in #667. Data result pruning faces two challenges:
1. It is highly inefficient especially at the early stages of pruning where most batches have not been pruned. We can avoid this problem by starting data result pruning once batch pruning catches up.
2. The main challenge of data result pruning is the lack of mapping from block height or batch number to data results. If we introduce a new map to facilitate data result pruning in the future, currently existing data results and new data results that support the new map should be distinguished correctly in queries, pruning, etc.
 
Here is a solution. We first add a new map at prefix `0x09` so we can simply prune data results belonging to a given batch when we prune it. We also store the last batch number at upgrade.

```
// Legacy data result store (To be pruned by batch pruning)
0x00 | is_batched | data_request_id | data_request_height -> data_result

// This batch assignment map can be shared.
0x01 | data_request_id | data_request_height              -> batch_number 

...

// New data result store (To be pruned by simple pruning)
// A data result object stored here has an entry at 0x09.
0x09 | is_batched | data_request_id | data_request_height -> data_result

// New map to support simple pruning 
0x10 | batch_number                                       -> data_request_id | data_request_height

// Last batch with legacy data results
0x11                                                      -> batch_number_at_upgrade
```

`BatchNumberAtUpgrade` is the batch number of the last batch containing legacy date results. This number would be set to `CurrentBatchNum - 1` in the upgrade handler. 

Data result storage and pruning would then follow this plan:

- After upgrade until last pruned batch number reaches `BatchNumberAtUpgrade`: 
  - Queries of data results should check both `0x00` and `0x09` maps.
  - Creation/update/deletion only uses the new map.
    - All data results belonging to batch with batch number up to `BatchNumberAtUpgrade` 
- Once the last pruned batch number reaches `BatchNumberAtUpgrade`: 
  - We activate pruning of old data results in the `0x00` map.  
  - Queries/updates/removals no longer check the `0x00` map.

hacheigriega:
## Motivation

Batching pruning addresses following issues:

1. We need to prune the accumulated batches and their associated data in batches.
2. Since legacy data results cannot be retrieved based on a batch number, they have to be pruned separately.
3. We need a permanent, efficient solution for batch pruning.

Issue 1 was implemented in #664. This PR addresses the other two issues.

### Pruning Parameters

Pruning logic is parameterized by the following module states, which will be added by the upgrade handler:

- `MaxBatchPrunePerBlock` is a module parameter that specifies the maximum number of batches to prune per block in the Batch Pruning strategy. (Defaults to 100)
- `MaxLegacyDataResultPrunePerBlock` is a module parameter that specifies the maximum number of legacy data results to prune per block in the Legacy Data Result Pruning strategy. (Defaults to 1000)
- `BatchNumberAtUpgrade` is the batch number of the latest batch at the time of upgrade. The upgrade handler should set the value in order to record the batch number of the last batch containing legacy data results.
- `HasPruningCaughtUp` is set to true when either of the following conditions is met:
    - (i) All batches up to batchNumberAtUpgrade have been pruned.
    - (ii) All batches up to `currentBatchNum - NumBatchesToKeep` have been pruned.
    
    It is set to false at the time of upgrade, initiating the Batch Pruning strategy. Once it is switched to true, the batching end blocker terminates the Batch Pruning strategy and initiates the Legacy Data Result Pruning strategy.

There is also a new global variable: 

- `NumBatchesToKeep` specifies the number of batches to preserve in the state. Currently set to 10000.
    - This value is not kept as a module state because it is a fixed value that originates from outside the module logic. In a future upgrade we may turn this into a module parameter, in which case the pruning logic would need to be updated to accommodate its variability.

### Pruning Strategies

**Basic Pruning (Implemented in this PR)**
When a new batch is created, the Basic Pruning strategy prunes one batch at `NewBatchNumber - NumBatchesToKeep`, if this number is higher than `BatchNumberAtUpgrade`.

We also temporarily add Batch Pruning and Legacy Data Result Pruning strategies to prune the batches and their data that have been accumulated until the upgrade.

**Batch Pruning (Implemented in #664)**
This pruning strategy prunes up to MaxBatchPrunePerBlock batches and their associated data, except for data results and their batch assignments, every block either up to batch at BatchNumAtUpgrade or up to batch at currentBatchNum - NumBatchesToKeep. Once either of the two termination conditions is met, HasPruningCaughtUp is switched to true, terminating Batch Pruning and initiating Legacy Data Result Pruning.

**Legacy Data Result Pruning (Implemented in this PR)**
Once `HasPruningCaughtUp` is switched to true, this pruning strategy begins. It iterates through legacy data result collection, pruning at most  `MaxLegacyDataResultPrunePerBlock` data results and their batch assignments per block.

## Related PRs and Issues

Closes #671


x/batching/keeper/batch_assignments.go
```diff
@@ -0,0 +1,67 @@
...
+			BatchNumber:       value,
+			DataRequestId:     key.K1(),
+			DataRequestHeight: key.K2(),
+		})
+		return false, nil
+	})
+	return batchAssignments, err
+}
+
+func (k Keeper) SetBatchDataResults(ctx context.Context, batchNumber uint64, dataRequestIDHeights types.DataRequestIDHeights) error {
```
Thomasvdam, line 57: Is this method used?

x/batching/keeper/pruning.go
```diff
@@ -0,0 +1,227 @@
...
+	}
+	err = k.batchSignatures.Clear(ctx, valRng)
+	if err != nil {
+		return err
+	}
+
+	dataResults, err := k.GetBatchDataResults(ctx, batchNumToPrune)
+	if err != nil {
+		if errors.Is(err, collections.ErrNotFound) {
+			k.Logger(ctx).Info("cannot prune batch because schema change has not been applied", "batch_num", batchNumToPrune)
```
Thomasvdam: This is not always true right? We can also have batches that only contain validator changes and no results.

Thomasvdam (commented):
Overall I think this looks good. I'm not sure if we did a double check of the proposed strategies and numbers with Mario.

x/batching/keeper/batch_assignments.go
hacheigriega: Explains the issue with data result pruning after upgrade 😅 Thank you.

x/batching/keeper/pruning.go
hacheigriega: Good catch.. We store an empty slice when constructing a batch so this does not give an error.

PR #683, opened 2026-02-25
Upgrade Cosmos SDK to v0.53

hacheigriega:
## Motivation

Cosmos SDK v0.50 is reaching the end of lifecycle at the end of Q1 2026 (see [here](https://github.com/cosmos/security/blob/main/POLICY.md)). Switching to the following "release family" will guarantee another year of support.
- CometBFT v0.38.x
- Cosmos SDK v0.53.x
- IBC v10.x

## Explanation of Changes

**CometBFT: No upgrade needed as we are already on v0.38.21**

**Cosmos SDK v0.50 -> v0.53**
- It features two new modules, x/epochs and x/protocolpool, which were not added in this PR. x/epochs is a periodic on-chain timer, and x/protocolpool is an improved community pool management module.
- x/crisis, along with its invariant checking, has been deprecated in favor of new simulator and x/circuit ([Source](https://github.com/cosmos/cosmos-sdk/pull/18380))
- Upgrade guidelines available [here](https://github.com/cosmos/cosmos-sdk/blob/main/UPGRADE_GUIDE.md#upgrade-handler).

**IBC v8 -> v10**
- This version's major feature is addition of IBC v2, which is not enabled in this PR as this is not the objective here. We may choose to enable this feature in the future.
- Beyond IBC v2, the key change is the removal of the following two modules:
  - Removal of the fee middleware module, which was used for relayer incentivization. I saw some discussions around migration of fee-enabled channels, but I was not able to find what conclusion they drew. Good news is that our chains do not have any fee-enabled channels (can be checked via `sedad q ibc-fee channels`), so we don't have to worry about this.
  - Removal of the capabilities module, which was used for authentication.
- Upgrade guidelines available [here](https://docs.cosmos.network/ibc/v10.1.x/migrations/v8_1-to-v10)

## Testing

Upgraded interchaintest to the latest version to pass the tests.

- Removed `TestChainUpgrade` altogether as it can be replaced by rehearsals we conduct before upgrades.
- Removed `TestConformance` and `TestSDKBoundaries` from CI due to the error `relative timeouts using block height is not supported`, which is caused by `strangelove-ventures/interchaintest/v8`'s [outdated usage](https://github.com/strangelove-ventures/interchaintest/blob/4bd4690ac4b89a8d73f5e3ea78158ea0ecaf653a/chain/cosmos/chain_node.go#L874-L876) of the `--packet-timeout-height` flag for the IBC transfer transaction. 
- Removed `TestSedaGaiaConformance` as it is covered by `TestConformance`.
- Fixed other tests that broke due to upgrade and added missing tests to the CI action.

Upgrade handler to be tested through a dry run of an upgrade.


x/staking/keeper/invariants.go
Thomasvdam, line 1: Do we need replacement functionality for this?

Thomasvdam (commented):
Do we need to take care of any state/params of the removed modules? Or is that done automatically?

x/staking/keeper/invariants.go
hacheigriega: Good points.. I am adding an upgrade handler that will remove x/crisis and x/capability. x/crisis apparently was not even working anyways (see [this](https://github.com/cosmos/cosmos-sdk/pull/18380)), and IBC v10 no longer uses x/capability for permission control. 

IBC also removes fee middleware, but we weren't using it anyways. I didn't see migration code for deleting this module, but that's probably because this module is stateless. I will double-check.

SDK team is suggesting use of its new simulator and x/circuit to replace this functionality, but I was thinking to create an issue and address that separately.

hacheigriega: Actually the fee middleware module also definitely has states, but there is no mention of how it is removed in the codebase or the docs. Checking other projects that performed similar upgrade..

Thomasvdam: Nice :D I think creating a separate issue is the right way to go, especially if it wasn't working!

app/upgrades/mainnet/v1.1.0/upgrade.go
```diff
@@ -0,0 +1,40 @@
+package v1
+
+import (
+	"context"
+
+	storetypes "cosmossdk.io/store/types"
+	upgradetypes "cosmossdk.io/x/upgrade/types"
+
+	"github.com/cosmos/cosmos-sdk/types/module"
+	crisistypes "github.com/cosmos/cosmos-sdk/x/crisis/types"
```
Thomasvdam, line 10: Why are we importing these types for the store key but hardcode the capability/ibcfee ones?

Thomasvdam (approved):
The simulation test did pass before? It failed because it's expecting a cosmos address but it should be seda. Did something in the config change?

app/upgrades/mainnet/v1.1.0/upgrade.go
hacheigriega: Ah, because they no longer exist in the new version we import since those modules have been removed

PR #685, opened 2026-03-09
Upgrade tallyvm to v2.6.1

hacheigriega:
We should merge this if we want to include this in our upcoming release.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -44,7 +44,7 @@ require (
 	github.com/pkg/errors v0.9.1
 	github.com/rakyll/statik v0.1.7
 	github.com/rs/zerolog v1.34.0
-	github.com/sedaprotocol/seda-wasm-vm/tallyvm/v2 v2.6.0
+	github.com/sedaprotocol/seda-wasm-vm/tallyvm/v2 v2.6.1
 	github.com/sedaprotocol/vrf-go v0.0.0-20231211075603-e5a17bb0b87c
 	github.com/spf13/cast v1.10.0
 	github.com/spf13/cobra v1.10.1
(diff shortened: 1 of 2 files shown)


mariocao (commented):
This is a breaking change, correct?

I see in the tallyvm that the get clock time function changed signature.
I think we don't use it in tally phases. 🤔

Thomasvdam:
> This is a breaking change, correct?
> 
> I see in the tallyvm that the get clock time function changed signature. I think we don't use it in tally phases. 🤔

We don't use it, but we need to provide a correct interface in the stub otherwise programs that use the import in the execution phase will be rejected in the tally, even though they're not using them. As a workaround we used separate programs for execution and tally.


Should we add this (which I think makes sense) we should make sure to test the upgrade on a node where we copied the wasm-vm directory (with all the compiled binaries) and execute a DR that was previously executed. I don't think we ever did a real test of the cache invalidation?

hacheigriega:
> Should we add this (which I think makes sense) we should make sure to test the upgrade on a node where we copied the wasm-vm directory (with all the compiled binaries) and execute a DR that was previously executed. I don't think we ever did a real test of the cache invalidation?

No we haven't done this testing. I think I am not fully understanding the problem. Why do we need this testing if we have been using programs without the import in question?

Thomasvdam:
> > Should we add this (which I think makes sense) we should make sure to test the upgrade on a node where we copied the wasm-vm directory (with all the compiled binaries) and execute a DR that was previously executed. I don't think we ever did a real test of the cache invalidation?
> 
> No we haven't done this testing. I think I am not fully understanding the problem. Why do we need this testing if we have been using programs without the import in question?

My bad. I meant testing an update to the wasm vm that invalidates the already cached Oracle Programs (so unrelated to the signature change 😄).

hacheigriega:
Thanks for pointing that out. Two-validator upgrade rehearsal using testnet export state (excluding the batches 😅) was successful. All IBC modules were migrated successfully, and an oracle program that was triggering the `clock_time_get: incompatible import type` tally error now gives a 252 exit code. A new cache directory is created instead under the updated wasm VM version. We should be good to release `v1.0.8-rc.1` once this is merged.

Thomasvdam (approved):

